PR #14, opened 2016-06-24
Update cli.rb to respect server version

rHorsey:


cli.rb
```diff
@@ -115,7 +115,11 @@ def find_or_create_target(target_type, aws_instance_options)
       puts 'Starting cluster...'
 
       # Don't use the old API (Version 1)
-      aws = OpenStudio::Aws::Aws.new
+      aws_options = {
+        ami_lookup_version: 2,
+        openstudio_server_version: options['openstudio_server_version']
```
shorowit: @rHorsey Seems like it actually needs to be `openstudio_server_version: aws_instance_options[:os_server_version]`. Can you please update?

PR #74, opened 2017-08-30
.tsvs for number of days to shift the draw profile with an even proba…

jmaguire1:
…bility for any number 0-364.


jmaguire1:
@shorowit this could use some review now that the draw profile changes in OpenStudio-BEopt have been merged in

PR #80, opened 2017-09-11
Utility bills test

joseph-robertson:


measures/UtilityBillCalculations/measure.rb
```diff
@@ -0,0 +1,351 @@
...
+  end
+  
+  def outputs
+    result = OpenStudio::Measure::OSOutputVector.new
+    result << OpenStudio::Measure::OSOutput.makeStringOutput("grid_cells")
+    result << OpenStudio::Measure::OSOutput.makeStringOutput("total_site_electricity")
+    buildstock_outputs = [
+                          "total_site_natural_gas",
+                          "total_site_propane",
+                          "total_site_other"
```
shorowit: I suggest removing the word "site" from the above 4 strings, as site/source doesn't have any meaning for utility bills.

joseph-robertson:
Yes I suppose the measure should be added to the other projects as well- I should probably spotcheck rates for the grid cells though first.

Also, can we split out propane/oil from "other" so we can apply rates to these fuels? How can this be done?

@ejhw

shorowit:
E+/OS only report "other" in their standard reporting mechanisms. See https://github.com/NREL/OpenStudio-BuildStock/issues/17 for a way to do it on our end, but it's hacky. Don't know if there any better approaches.

PR #94, opened 2017-09-23
Remove non-linux binaries from sam sdk in photovoltaics measure.

joseph-robertson:


Rakefile
```diff
@@ -313,6 +323,7 @@ def get_all_project_dir_names()
     Dir.entries(File.dirname(__FILE__)).each do |entry|
         next if not Dir.exist?(entry)
         next if not entry.start_with?("project_")
+        next if entry.end_with?("efs")
```
shorowit: Why exclude this project?

joseph-robertson: Ugh I forgot to take that out. (Was temporarily excluding it because it's not yet committed to the master branch.)

PR #109, opened 2017-10-12
EnergyPlus measure for utilizing RunPeriod:CustomRange object

joseph-robertson:


joseph-robertson:
I tested this measure on liquidcool and it appears to be working.

PR #142, opened 2018-02-27
Spot ventilation

jmaguire1:
This branch allows the spot ventilation for bathrooms, range, and clothes dryer to be an input to the airflow measure. For ResStock runs there's probability distributions based on the occupancy, range, and clothes dryer schedule respectively that are used to pick the hour of spot ventilation. I've done a fair bit of testing that the spot ventilation is being assigned to the correct hours and that the sampling is done correctly.


jmaguire1:
I'll fix that when it gets merged into master.

PR #152, opened 2018-06-19
Downscoping in PAT

joseph-robertson:


measures/BuildExistingModel/measure.rb
```diff
@@ -89,18 +84,26 @@ def run(model, runner, user_arguments)
...
       runner.registerInfo("Generating buildstock.csv sampling results.")
-    
-      require 'json'
-      # analysis = JSON.parse(File.absolute_path(File.join(File.dirname(__FILE__), "..", "..", "analysis.json")))
-      # runner.registerInfo(analysis)
-      num_datapoints = 100 # TODO: NUMDATAPOINTS=`awk -F\"maximum\": 'NF>=2 {print $2}' analysis.json | sed 's/,//g' | head -n1 | xargs` # Yes, this is gross.
+
+      num_datapoints = nil
+      File.readlines(File.absolute_path(File.join(File.dirname(__FILE__), "..", "..", "analysis.json"))).each do |line|
+        if matches = line.match(%r{\s+"maximum":\s+(?<num_datapoints>\d+)})
```
shorowit: Super gross. Why not use the ruby json library? (I had to use awk/sed/etc because I was parsing from a bash script.)

joseph-robertson: I originally tried that, but got an "unrecognized token" or something. I'll try again.

shorowit: Weird. That might suggest it's a malformed json. If that's the case, we should at least get it fixed upstream.

joseph-robertson:
I still need to move the sampling back out of the measure and into the worker init script.

joseph-robertson:
@shorowit Are you good with merging this into master?

shorowit:
I will plan to review.

PR #157, opened 2018-08-06
Latest measures, fixes to options lookup for no heating systems, and …

joseph-robertson:
…info statements in sampling scripts.


shorowit:
@joseph-robertson  Why did you associate the ResidentialAirflow has_hvac_flue argument with the Heating Fuel parameter and not the HVAC System Heating Foo parameters?

joseph-robertson:
@shorowit Seems like since every sample is assigned a value for all HVAC System Heating Foo parameters, the has_hvac_flue argument would be assigned multiple times?

shorowit:
Ah, I think I understand. Clever.

PR #159, opened 2018-09-12
Pointing to latest default ami.

joseph-robertson:


joseph-robertson:
Maybe my understanding of permalink was wrong. https://stackoverflow.com/questions/23821235/how-to-link-to-specific-line-number-on-github.

shorowit:
Right, it's a permalink. So when we update to 2.6.1, the link will still be pointing to 2.6.0.

joseph-robertson:
I interpreted it as permanently linking to that portion of code in all future versions of the file.

shorowit:
That's not right. Your link is pointing to a specific commit so it will never change. Someone clicking on it will also see that exact line in code from that version of the file. Hence why it's called a permalink.

Even if you did link to the lines of code that updated with new versions of the file (the link would be: https://github.com/NREL/OpenStudio-BuildStock/blob/master/project_resstock_national/pat.json#L49), there's no guarantee that the line number will stay the same.

Long story short, there's no way to accomplish what you want.

joseph-robertson:
@ejhw

ejhw:
Ah sorry, I was wrong about how permalinks work. Maybe we should just update the ami name as necessary.

PR #187, opened 2018-12-07
Add note on manual sampling script downselecting

ejhw:


ejhw:
Thanks, good updates.

PR #195, opened 2019-01-03
Fix lighting distribution in national project

ejhw:
Old numbers were caused by misinterpretation of RECS data and were corrected in post-processing for the QER analysis. This PR updates the distributions to the correct RECS 2009 distributions.


ejhw:
Good idea. Let's touch base with @mpathak24 on updating the RECS query script, but let's merge for now.

PR #199, opened 2019-01-16
Download datapoints dependency missing from gemfile

joseph-robertson:
Spotted by Philip.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,6 @@ source 'http://rubygems.org'
 gem 'rake', '~> 11.2.2'
 gem 'colored', '~> 1.2'
 gem 'git', require: false
-gem 'rubyzip'
 
 group :test do
   gem 'minitest', '~> 5.9'
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,7 @@ source 'http://rubygems.org'
 gem 'rake', '~> 11.2.2'
 gem 'colored', '~> 1.2'
 gem 'git', require: false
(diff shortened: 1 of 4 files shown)


shorowit:
Wouldn't it be better to use [OpenStudio's UnzipFile class](https://openstudio-sdk-documentation.s3.amazonaws.com/cpp/OpenStudio-2.7.1-doc/utilities/html/classopenstudio_1_1_unzip_file.html)?

joseph-robertson:
Yeah probably. I think I copied that portion of code from one of Ry's scripts.

joseph-robertson:
@shorowit OK how about now?

joseph-robertson:
@shorowit If we use OpenStudio's UnzipFile class, wouldn't we have to introduce a dependency on `openstudio` in the `download_datapoints.rb` script? Do we want that?

shorowit:
The script can easily be run via `openstudio.exe download_datapoints.rb` if it had a `require openstudio` in it, I believe. (And of course the openstudio dependency only needs to be referenced if the user chooses to call the script with the unzip argument.)

shorowit:
If user the specifies the unzip argument, you can wrap the `require openstudio` in a begin/rescue and then show the user a nice error msg that they need to call the script with the openstudio executable.

joseph-robertson:
How about now? I just updated the docs to instruct the user to run the script using the openstudio cli all the time. That way I could leave out the `require 'openstudio'` altogether.

PR #210, opened 2019-02-14
Pumps disaggregation

joseph-robertson:


joseph-robertson:
I think I still need to meter dehumidifiers as cooling electricity. But not sure why the testing project didn't catch it...

PR #211, opened 2019-02-20
Dependency graphs and wheels

afontani:
I put together scripts in the docs folder to automatically generate dependency graphs and a dependency wheel for each <project_folder> in the OpenStudio-BuildStock repository.  These scripts create a util folder if it doesn't exist and puts data for these visualizations.  A project README.md file has been added to each <project_folder> where there is a link to the interactive dependency wheel.  These visualizations are easily updated as they depend strictly on the TSV files.  They are updated through a single regenerate_visualization.ipynb script.


docs/visualizations/README.md
```diff
@@ -0,0 +1,35 @@
+# Create visualizations
```
ejhw: Mention that the files in this folder (e.g., OpenStudio-BuildStock/docs/visualizations/dependencyWheels/dep_wheel.html) are blank templates so users aren't confused as to why they are blank. Would be good to mention that in the template text itself too.

afontani: I added a commit to make this more explicit. There is now language in the README.md and the file has been renamed.

PR #234, opened 2019-04-16
Integrity check changes

joseph-robertson:
@afontani ran into an issue where the integrity check for his custom project failed due to memory error (specifically on this line: https://github.com/NREL/OpenStudio-BuildStock/blob/master/Rakefile#L340).

We believe this happens because that line is trying to take the product of ALL options in the options lookup, instead of just those options found in the project you are running the integrity check against.

@shorowit

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -329,7 +329,7 @@ def integrity_check_options_lookup_tsv(project_dir_name)
     param_names = measures[measure_subdir].keys()
     options_array = []
     param_names.each do |parameter_name|
-      if ["Bathroom Spot Vent Hour", "Clothes Dryer Spot Vent Hour", "Range Spot Vent Hour"].include? parameter_name
+      if ["Bathroom Spot Vent Hour", "Range Spot Vent Hour"].include? parameter_name
         # Prevent "too big to product" error for airflow measure by just
         # using first option for these parameters.
         options_array << [measures[measure_subdir][parameter_name].keys()[0]]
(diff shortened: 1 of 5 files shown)


joseph-robertson:
I feel like there is still a scenario that goes uncheck.

Consider this example in options_lookup.tsv:

parameter | option | measure
-- | -- | --
Cooking Range | Gas, 80% Usage | ResidentialApplianceCookingRange
Cooking Range | Gas, 100% Usage | ResidentialApplianceCookingRange
Cooking Range | Gas, 120% Usage | ResidentialApplianceCookingRange
Cooking Range | None |  
Cooking Range Schedule | California Average | ResidentialApplianceCookingRange
Cooking Range Schedule | None |  

Say Cooking Range.tsv has an option for None but Cooking Range Schedule.tsv does not. Then BuildExistingModel can fail on building the model since not all the cooking range measure's arguments exist. However, the integrity check passes because None isn't included in these keys: https://github.com/NREL/OpenStudio-BuildStock/blob/master/Rakefile#L337. It appears to only check combinations of non-"None" options.

@shorowit

shorowit:
It intentionally checks all options in options_lookup.tsv, not just those specified in the project tsv's, because any option could be specified by the user as an upgrade.

shorowit:
This PR is very confusing because it seems like the proposed fix is a [one-liner](https://github.com/NREL/OpenStudio-BuildStock/pull/234/files#diff-52c976fc38ed2b4e3b1192f8a8e24cffR291) and then there are a bunch of other changes thrown in?

shorowit:
Regarding an unchecked scenario, I'm about to open a new PR that adds unit tests for the integrity checks. You can add the scenario there.

joseph-robertson:
It would've been a one-liner, but the three geometry measure issue required that the national project (which only assigns arguments for one out of three of the geometry measures) remove keys for the sfa and mf geometry measures. The other tsv changes were just unrelated simplifications.

joseph-robertson:
If this approach doesn't work, we should talk with Tony and figure out what approach will work.

PR #237, opened 2019-04-19
Integrity check unit tests

shorowit:
Adds unit tests to make sure that the integrity checks are covering various potential scenarios that would cause errors. Each unit test consists of a housing_characteristics dir with custom TSVs and corresponding options in the test_options_lookup.tsv that should cause the error. The unit tests check that the appropriate error message is hit.

I took a first cut at adding various unit tests:

TSV files:
- Non-numeric value
- Duplicate rows
- Missing parent
- Missing row
- Row doesn't sum to 1
- Unused file

options_lookup.tsv:
- Bad measure argument value
- Extra measure argument
- Missing measure argument
- Missing measure


shorowit:
The new tests can be run with `rake integrity_check_unit_tests`.

Issue #235, opened 2019-04-18
Update outages measure


joseph-robertson:
With removal of the "10000 setpoint workaround", I'm not sure if the outages measure is behaving as expected.

shorowit:
This was brought up during the setpoints PR. I don’t understand why it was merged without addressing this.

joseph-robertson:
I'm going to look into this.

PR #238, opened 2019-04-19
Updates to outage measure

joseph-robertson:
#235 

It looked like the last hour of the day wasn't getting the new schedule applied. This fixes that. I've also cleaned up some code and added some new unit tests.

@jmaguire1 can you review this?

Specifically, it looks like some things aren't getting zeroed out (e.g., electric interior equipment). You can use the attached osw file for testing.
[example_single_family_detached.osw.txt](https://github.com/NREL/OpenStudio-BuildStock/files/3099490/example_single_family_detached.osw.txt)


joseph-robertson:
We've removed our `Constants.NoHeatingSetpoint` and `Constants.NoCoolingSetpoint` values, and so I threw in `-10000` and `10000` here thinking it would give the same result as before. Not really sure what you mean by actuate an availability schedule. Let's chat when you have time.

joseph-robertson:
I'm now setting an "outage" fixed interval schedule (of 0s and 1s) on availability schedules for HVAC equipment and EMS sensor res_infil_1_wh_sch_s (for the house fan).

This appears to mostly work; however there is still tiny amounts of electric heating during the outage period. Any idea what this could be?

@jmaguire1

PR #239, opened 2019-04-19
Integrity check performance improvement

shorowit:
Dramatically improves the speed of performing measure argument checks. Rather than checking every combination of option for every parameter that contributes to a single measure's arguments, we now pick options from each parameter in step.

This should not be any less robust. Rather, the previous code was written for simplicity and did not imagine a future where the TSVs/measures were organized as they currently are.

Suppose we have these parameter/options that define the arguments for a measure:

Parameter 1 | Parameter 2
-- | --
Option 1A | Option 2A
Option 1B | Option 2B
Option 1C | Option 2C
Option 1D |  

Previously, we validated measure arguments for all 12 option combinations: 

1. Option 1A, Option 2A
2. Option 1B, Option 2A
3. Option 1C, Option 2A
4. Option 1D, Option 2A
5. Option 1A, Option 2B
6. Option 1B, Option 2B
7. Option 1C, Option 2B
8. Option 1D, Option 2B
9. Option 1A, Option 2C
10. Option 1B, Option 2C
11. Option 1C, Option 2C
12. Option 1D, Option 2C

Now, we only check 4 option combinations:

1. Option 1A, Option 2A
2. Option 1B, Option 2B
3. Option 1C, Option 2C
4. Option 1D, Option 2A

The more options/parameters that contribute to a measure's arguments, the larger the reduction in runtime speed.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -317,7 +317,6 @@ def integrity_check_options_lookup_tsv(project_dir_name)
     end
   end
 
-  max_checks = 1000
   measures.keys.each do |measure_subdir|
     puts "Checking for issues with #{measure_subdir} measure..."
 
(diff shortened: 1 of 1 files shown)


shorowit:
This should address @afontani's issue.

shorowit:
Integrity checks according to the CI:

  | master (sec) | this branch (sec) | runtime improvement
-- | -- | -- | --
project_resstock_testing | 47.1 | 28.4 | 39.7%
project_resstock_multifamily | 93.3 | 71.8 | 23.1%
project_resstock_national | 46.9 | 30.8 | 34.3%

shorowit:
Can you imagine situations where, say, the combinations "Option 1A + Option 2A" and "Option 1B + Option 2B" have complete arguments specified, but the combination "Option 1A + Option 2B" (only checked in old approach) does not?

PR #245, opened 2019-04-30
Register climate zones

joseph-robertson:
In response to:
* Chuck
* Ed

This is a pretty simple and straightforward PR: it adds two columns "climate_zone_ba" and "climate_zone_iecc" (based on the epw) to the results csv.

@ejhw


shorowit:
It should work for either assuming the weather station WMO has been set appropriately in the EPW weather file's header.

PR #263, opened 2019-05-31
HVAC autosizing and add/replace fixes

joseph-robertson:
* When a model has both a central system and non central system, don't autosize the non central system (e.g., central boiler with room ac)
* Cannot have heating-only fan coil anymore; the ZoneHVACUnitHeater object was not being autosized correctly and was resulting in zero heating energy
* Don't remove the cooling-only fan coil when applying a heating-only non central system (e.g., cooling-only fan coil with furnace)


joseph-robertson:
* BoilerBaseboards/FanCoilHeatingOnly + CentralAC/RoomAC: cooling comes way down, especially for CentralAC

joseph-robertson:
* Appears to be a bug where FanCoilHeatingOnly consumes no energy. I'm planning to send Jeff and idf to investigate.

joseph-robertson:
Regression tests:
* [master.csv.txt](https://github.com/NREL/OpenStudio-BuildStock/files/3243062/master.csv.txt)
* [autosize-fix.csv.txt](https://github.com/NREL/OpenStudio-BuildStock/files/3243063/autosize-fix.csv.txt)

PR #268, opened 2019-06-14
fix: HVAC System Cooling Type - added dependency

mpathak24:
HVAC System Cooling Type.tsv structure changed; "Is heat pump" dependency added.


ejhw:
@mpathak24 Check integrity and resolve conflicts with master

PR #271, opened 2019-06-17
Insulation finished roof; structural changes

mpathak24:


project_singlefamilydetached/housing_characteristics/Insulation Finished Roof.tsv
```diff
@@ -1,2 +1,2 @@
-Option=Uninsulated	Option=R-30
-0.5	0.5
```
joseph-robertson: Were we previously simulating any SFD houses with finished roofs? Anyhow, seems like you don't need the `Dependency=Geometry Building Type` column here since there's only one enumeration in the list.

mpathak24: This TSV was added after the national SFD project was over; but since the mf project this has been the case and I was directed to add the dependency and set-it up as "None" for all SFs. I guess we can remove the TSV all together from this project?

joseph-robertson: We can't remove "Insulation Finished Roof.tsv" from project_singlefamilydetached (see https://trello.com/c/N5xhoBQL/126-why-is-insulation-finished-rooftsv-in-all-projects) unless we do something with "Roof Material.tsv" (e.g., split into separate "Unfinished Roof Material.tsv" and "Finished Roof Material.tsv"). My point was just that you don't need column A (`Dependency=Geometry Building Type`) in this file.

ejhw: As long as `Insulation Finished Roof.tsv` is in the project, then it will get sampled and show up in `results.csv`, right (even if it isn't used)? Adding `Dependency=Geometry Building Type` will at least allow us to set `Insulation Finished Roof` to `None` for SFD homes, so it seems like a good idea to me.

shorowit: There's only one option (None) set at 100%, so the dependency has no impact. You'll always get None in the results.csv with or without the dependency.

mpathak24: So what's the direction for going ahead here? remove the dependency or keep it?

shorowit: Remove the dependency. It's useless.

ejhw:
Resolve conflicts and merge

mpathak24:
I discussed with @shorowit, and @joseph-robertson about the error caused by the "None" option in the Insulation finished roof.tsv because of the missing required values for the arguments for the `ResidentialConstructionsFinishedRoof` such as `cavity_r`,  `cavity_depth` etc. I and @joseph-robertson were leaning towards either a) separate the Roof Material for finished and unfinished roof and add the dependency of building type such that only finished roof of MF building types has roof material properties associated with it. @shorowit suggested an easy "way-around" this issue, i.e., assign the `None` option the same features as the `Uninsulated` option. By opting for this approach, the SF building type will have properties assigned for the finished roof but no surfaces to assign these properties to and hence no finished roof.

ejhw:
> I discussed with @shorowit, and @joseph-robertson about the error caused by the "None" option in the Insulation finished roof.tsv because of the missing required values for the arguments for the `ResidentialConstructionsFinishedRoof` such as `cavity_r`, `cavity_depth` etc. I and @joseph-robertson were leaning towards either a) separate the Roof Material for finished and unfinished roof and add the dependency of building type such that only finished roof of MF building types has roof material properties associated with it. @shorowit suggested an easy "way-around" this issue, i.e., assign the `None` option the same features as the `Uninsulated` option. By opting for this approach, the SF building type will have properties assigned for the finished roof but no surfaces to assign these properties to and hence no finished roof.

I like @shorowit 's suggestion.

shorowit:
So you ended up going a different approach? https://github.com/NREL/OpenStudio-BuildStock/pull/271/commits/7c3228fd4eb8ff2f7f8402cb29b056563056ef68

I guess it's fine, although it won't work if we ever start modeling single-family detached homes with, e.g., conditioned attics, cathedral ceilings, or flat roofs.

joseph-robertson:
If we ever did start modeling SFD with conditioned attics, it'd probably be using the new HPXML workflow anyway, right? @ejhw?

ejhw:
> If we ever did start modeling SFD with conditioned attics, it'd probably be using the new HPXML workflow anyway, right? @ejhw?

Yes, I suppose so. Though it is tempting to try to start using it now since there is an existing `ResidentialConstructionsFinishedRoof` measure and we have the data for it ready.

joseph-robertson:
We have the data for it? For SFD homes?

PR #280, opened 2019-06-27
Update testing project to sweep thru more options

joseph-robertson:


joseph-robertson:
The integrity check for the testing project would throw an error if rows didn't sum to 1. So I think you're dividing by the right numbers.

PR #287, opened 2019-07-08
Add year argument to simulation controls measure

joseph-robertson:
https://trello.com/c/XLUQN9WK/146-add-year-argument-to-simulation-controls-measure


shorowit:
I think some more description is needed here. What happens if you provide a year and use AMY weather?

joseph-robertson:
> I think some more description is needed here. What happens if you provide a year and use AMY weather?

https://github.com/NREL/OpenStudio-BuildStock/blob/year-arg/measures/ResidentialSimulationControls/measure.rb#L135

shorowit:
This needs to be documented up front (argument description, RTD), not as an after-the-fact info message.

PR #293, opened 2019-07-15
Increase upgrade options from 20 to 25

joseph-robertson:
https://trello.com/c/X7b5iMbG/137-increase-num-upgrades-to-25


docs/read_the_docs/source/upgrade_scenario_config.rst
```diff
@@ -49,6 +49,14 @@ Or say we want to apply the upgrade only to houses with 3 car garages that aren'
 .. todo::
    
    Come up with some better examples here.
+   
+Currently, you can enter up to 25 options per upgrade. To allow additional options per upgrade, you would need to increase the value returned by the following method defined in ``resources/HPXMLtoOpenStudio/resources/constants.rb``:
```
shorowit: Good idea, but I think they will also have to update measure.xmls (and the outputs section of the PAT projects)? Maybe language describing how to do this should go in the advanced tutorial section.

project_testing/measures/SimulationOutputReport/measure.rb
```diff
@@ -90,49 +98,14 @@ def outputs
...
-      "upgrade_option_19_lifetime_yrs",
-      "upgrade_option_20_cost_usd",
-      "upgrade_option_20_lifetime_yrs",
-      "weight"
+      "upgrade_cost_usd"
     ]
+    for option_num in 1..num_options
+      buildstock_outputs << "upgrade_option_%02d_cost_usd" % option_num
+      buildstock_outputs << "upgrade_option_%02d_lifetime_yrs" % option_num
+    end
```
shorowit: 👍

shorowit: I assume it's not a big deal that you're changing the order of the outputs (and, presumably, results.csv columns)?

joseph-robertson: How is this changing the order of the outputs?

shorowit: It looks like "upgrade_cost_usd" used to be last in the list (after "weight"), now it's before all the "upgrade_option_foo"?

shorowit: Oh wait, I think I'm reading the changes wrong. Maybe it is the same.

shorowit: Nevermind, it looks good.

joseph-robertson: Aren't they before? https://github.com/NREL/OpenStudio-BuildStock/blob/master/measures/SimulationOutputReport/measure.rb#L92-L93

PR #301, opened 2019-08-07
Remove shared facades

joseph-robertson:
https://trello.com/c/PAKSLk6c/141-review-shared-facades-tsv


rajeee:
On quick comparison with master, this change introduced only about +1% annual energy consumption.

PR #305, opened 2019-08-15
Low flow fixture options

joseph-robertson:


resources/options_lookup.tsv
```diff
@@ -1153,6 +1153,8 @@ Water Heater	FIXME Other Fuel	ResidentialHotWaterHeaterTank	fuel_type=gas	tank_v
 Hot Water Fixtures	50%	ResidentialHotWaterFixtures	shower_mult=0.5	sink_mult=0.5	bath_mult=0.5	plant_loop=auto																																												
 Hot Water Fixtures	100%	ResidentialHotWaterFixtures	shower_mult=1.0	sink_mult=1.0	bath_mult=1.0	plant_loop=auto																																												
 Hot Water Fixtures	200%	ResidentialHotWaterFixtures	shower_mult=2.0	sink_mult=2.0	bath_mult=2.0	plant_loop=auto																																												
+Hot Water Fixtures	Low Flow 90%	ResidentialHotWaterFixtures	shower_mult=0.9	sink_mult=0.9	bath_mult=1.0	plant_loop=auto																																												
+Hot Water Fixtures	Low Flow 95%	ResidentialHotWaterFixtures	shower_mult=0.95	sink_mult=0.95	bath_mult=1.0	plant_loop=auto
```
shorowit: The names are pretty confusing, because they are mixing occupant usage (50%, 100%, 200%) with fixture types (low flow or not). Maybe you can make the names more consistent? How about:
- 50% Usage
- 100% Usage
- 100% Usage, Low Flow
- 200% Usage

And not sure why you need two different low flow options, seems like you should choose the more appropriate number...

joseph-robertson: Hmm. If we're going for consistency, I think it should be more like "95% Usage (Low Flow)".

Is there a universally accepted "low flow" value?

shorowit: I don't think you're following. Does this make it clearer?

Option Name | Multiplier
-- | --
50% Usage | 0.50
50% Usage, Low Flow | 0.48
100% Usage | 1.00
100% Usage, Low Flow | 0.95
200% Usage | 2.00
200% Usage, Low Flow | 1.90

The two options you created apply the low-flow fixtures to the 100% Usage option.

jmaguire1: Agreed this is an issue, but Joe and I were having a conversation earlier about what number to use for "low flow" fixtures. I remember we used to have them as a specific option in BEopt a while ago and eventually removed them because I think there was some conflicting results about how much water you actually save with low flow fixtures. 95% is coming straight from Addendum A from RESNET, but if there are better places to pull a number from that would be helpful to know.

resources/options_lookup.tsv
```diff
@@ -1150,9 +1150,12 @@ Water Heater	"Electric Heat Pump, 80 gal"	ResidentialHotWaterHeaterHeatPump	stor
...
 Water Heater	FIXME Other Fuel	ResidentialHotWaterHeaterTank	fuel_type=gas	tank_volume=auto	setpoint_temp=125	location=auto	capacity=40	energy_factor=0.59	recovery_efficiency=0.76	offcyc_power=0	oncyc_power=0																																							
 		ResidentialAirflow	has_water_heater_flue=false																																															
-Hot Water Fixtures	50%	ResidentialHotWaterFixtures	shower_mult=0.5	sink_mult=0.5	bath_mult=0.5	plant_loop=auto																																												
-Hot Water Fixtures	100%	ResidentialHotWaterFixtures	shower_mult=1.0	sink_mult=1.0	bath_mult=1.0	plant_loop=auto																																												
-Hot Water Fixtures	200%	ResidentialHotWaterFixtures	shower_mult=2.0	sink_mult=2.0	bath_mult=2.0	plant_loop=auto																																												
+Hot Water Fixtures	50% Usage	ResidentialHotWaterFixtures	shower_mult=0.5	sink_mult=0.5	bath_mult=0.5	plant_loop=auto																																												
+Hot Water Fixtures	"50% Usage, Low Flow"	ResidentialHotWaterFixtures	shower_mult=0.48	sink_mult=0.48	bath_mult=0.5	plant_loop=auto
```
jmaguire1: I don't know if it will matter, but why are there quotes only around the low flow options? If you could just remove those I think everything looks fine

joseph-robertson: When you save in Excel, it slaps quotes around any options with a comma in it. Annoying, but doesn't matter.

Issue #306, opened 2019-08-19
setUseSideDesignFlowRate for solar hw measure


joseph-robertson:
`storage_tank.setUseSideDesignFlowRate(UnitConversions.convert(shw_system.storage_vol, "ft^3", "m^3") / 60.1)`

PR #307, opened 2019-08-19
Addresses #306

joseph-robertson:


resources/measures/ResidentialHotWaterSolar/measure.rb
```diff
@@ -415,6 +415,7 @@ def run(model, runner, user_arguments)
         storage_tank.setNode1AdditionalLossCoefficient(0)
         storage_tank.setNode6AdditionalLossCoefficient(0)
         storage_tank.setSourceSideDesignFlowRate(UnitConversions.convert(shw_system.coll_flow, "cfm", "m^3/s"))
+        storage_tank.setUseSideDesignFlowRate((UnitConversions.convert(shw_system.storage_vol, "ft^3", "m^3")) / 60.1)
```
jmaguire1: Could you add a comment here (and maybe in the HPWH measure too) that the 60.1 here is to ensure that autosizing doesn't lead to the entire tank volume being drawn in one minute?

joseph-robertson: Yep: ` # 60.1 (seconds) ensures that the autosizing doesn't draw the whole tank volume in one minute`

PR #313, opened 2019-08-23
Adds integrity checks for names with illegal characters

shorowit:
Integrity checks now look for illegal characters (`(`, `)`, `&`, and `|`) in either parameter or option names in the options_lookup.tsv.


shorowit:
There's at least one bad name in the current set of options which will cause the CI to fail with these new tests. Can someone address any issues that are found? @joseph-robertson @vtnate

PR #315, opened 2019-08-29
Log the error string along with backtrace

rajeee:
Currently, the apply_measure function only logs the error backtrace when the measure being applied fails, but not the actual error message. This fixes the issue, and logs the error message alongside the backtrace.


joseph-robertson:
@rajeee I defer to @shorowit.

PR #361, opened 2020-01-14
HVAC-GEB

jmaguire1:
## Pull Request Description

Adding in additional functionality for modeling HVAC equipment more realistically for GEB scenarios. High level additions for this branch include:

- [ ] Add optional deadband argument for thermostats/deadbands
- [ ] Account for cycling behavior when using realistic thermostats
- [ ] Add realistic 2 speed HVAC performance
- [ ] Allow max PLR to be set for fully variable speed equipment
- [ ] Investigate thermal mass/mixing assumptions affecting behavior

## Checklist
- [ ] PAT project measures/outputs have been updated


resources/measures/ResidentialHVACCoolingSetpoints/measure.rb
```diff
@@ -116,8 +116,14 @@ def arguments(model)
     season_end_month.setDescription("End month of the cooling season.")
     season_end_month.setDefaultValue("Dec")
     args << season_end_month
-    
-    #TODO: Add thermostat deadband argument
+
+    # Make a string argument for thermostat deadband
+    onoff_thermostat_deadband = OpenStudio::Measure::OSArgument::makeStringArgument("onoff_thermostat_deadband", true)
```
shorowit: Is there a reason to use a string argument instead of a double argument?

resources/measures/ResidentialHVACHeatingSetpoints/measure.rb
```diff
@@ -116,8 +116,14 @@ def arguments(model)
...
-    
-    #TODO: Add thermostat deadband argument
+
+    # Make a string argument for thermostat deadband
+    onoff_thermostat_deadband = OpenStudio::Measure::OSArgument::makeStringArgument("onoff_thermostat_deadband", true)
+    onoff_thermostat_deadband.setDisplayName("OnOff Thermostat Deadband")
+    onoff_thermostat_deadband.setDescription("Specify a positive value of deadband between setpoint and cutout temperature to model on/off thermostat behavior. The sama value applies to both heating and cooling. Default is 0.0(traditional E+ thermostat).")
+    onoff_thermostat_deadband.setUnits("degrees F")
+    onoff_thermostat_deadband.setDefaultValue("0.0")
+    args << onoff_thermostat_deadband
```
shorowit: What happens if the heating and cooling setpoint measures specify different deadband values? It seems like whichever measure runs last will currently override the other one.

yzhou601: True, I was also concerning this issue. I think the best approach is just adding one argument for the whole workflow. However, the measures are split into heating and cooling parts (if not creating a new measure). And the specified value can be 0.0 for either (which is also default) which cannot tell whether the other one has specified input or just kept it as default. Therefore I ended up clarifying it in the description of argument that this value will be applied to both heating and cooling thermostat. Might have missed something because I am not very familiar with this repo yet.

shorowit: How about throwing an error if the deadband has already been set and it's now being set to a different value than before? That would ensure a single value is used.

yzhou601: Eg. if someone specifies cooling deadband = 0.0 and heating deadband = 1, it will not be fairly detected as an error because 0.0 is also the default value when there's nothing specified (is_initialized not available in this case). But we might want to capture this to let people know two deadbands are not supported. Though not sure how often this might occur.

yzhou601: Oh sorry if I confused you.

yzhou601: Sorry I was thinking of 0 as valid input, maybe it can just be ignored.

yzhou601: Error check added

resources/measures/ResidentialHVACCoolingSetpoints/measure.rb
yzhou601: Resolved

workflows/example_single_family_detached_TMY.osw
```diff
@@ -248,8 +247,7 @@
         "weekday_setpoint": "76.0",
         "weekend_offset_magnitude": "0",
         "weekend_offset_schedule": "0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0",
-        "weekend_setpoint": "76.0",
-        "onoff_thermostat_deadband": "0"
```
yzhou601: @jmaguire1 It seemed these example osws already auto-updated(not covered by my manual changes, sorry for misleading) their arguments to include the new thermostat deadband (default to 0).

jmaguire1:
We're going to do this in the HPXML branch instead, see https://github.com/NREL/OpenStudio-HPXML/pull/754

PR #372, opened 2020-01-23
Fixes wood stove upgrade

shorowit:
Reported at https://unmethours.com/question/42433/some-upgrade-options-arent-compatible-with-resstock/

I added a bugfix for the upgrade so that the EnergyPlus simulation is successful. 

However, the upgrade is still failing because the SimulationOutputReport measure isn't looking at the wood fuel (OtherFuel1). @joseph-robertson, can you update the measure? We only allow wood for space heating in ResStock, so you don't have to handle other end uses. FWIW, here's the error message from run.log:
`[16:06:31.407796 ERROR] Disaggregated total site energy (31.626833704418196 GJ) relative to building total site energy (37.77 GJ): -6.143166295581807 GJ.`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ Features
 - Separate tsv files for bedrooms, cooking range schedule, corridor, holiday lighting, interior/other lighting use, pool schedule, plug loads schedule, and refrigeration schedule ([#338](https://github.com/NREL/OpenStudio-BuildStock/pull/338))
 
 Fixes
-- Account for wood heating energy in simulation output ([#372](https://github.com/NREL/OpenStudio-BuildStock/pull/372))
+- Allow Wood Stove option as an upgrade, and acccount for wood heating energy in simulation output ([#372](https://github.com/NREL/OpenStudio-BuildStock/pull/372))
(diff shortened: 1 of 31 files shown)


measures/SimulationOutputReport/measure.rb
```diff
@@ -20,7 +20,7 @@ def name
   end
 
   def description
-    return "Reports simulation outputs of interest."
+    return "Reports simulation ou tputs of interest."
```
shorowit: ☝️

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ Features
 - Separate tsv files for bedrooms, cooking range schedule, corridor, holiday lighting, interior/other lighting use, pool schedule, plug loads schedule, and refrigeration schedule ([#338](https://github.com/NREL/OpenStudio-BuildStock/pull/338))
 
 Fixes
+- Account for wood heating energy in simulation output ([#372](https://github.com/NREL/OpenStudio-BuildStock/pull/372))
```
shorowit: Also fixes allowing the Wood Stove option as an upgrade?

Issue #380, opened 2020-01-31
Options_lookup assigns the same argument multiple times


yzhou601:
See following an example for *Heating/Cooling Setpoint Offset Magnitude* argument:

![image](https://user-images.githubusercontent.com/32500275/73578452-e772ac00-443c-11ea-948f-b6a4dc634571.png)

PR #381, opened 2020-01-31
Addresses #380, removing multiple setpoint arguments from options lookup

joseph-robertson:
Addresses #380.

## Pull Request Description

[description here]

## Checklist

Not all may apply:

- [ ] ~PAT project measures/outputs have been updated~
- [ ] ~Documentation has been modified appropriately~
- [ ] ~Any new options are added to `project_testing`~
- [ ] ~`project_testing` runs without any failures~
- [ ] ~The [changelog](https://github.com/NREL/OpenStudio-BuildStock/blob/master/CHANGELOG.md) has been updated appropriately~


shorowit:
It seems like the integrity checks should be updated to catch this.

resources/options_lookup.tsv
```diff
@@ -1224,388 +1224,267 @@ HVAC System Is Shared	Only Cooling Shared
...
 		ResidentialAirflow	has_hvac_flue=true																																												
-		ResidentialHVACSizing	show_debug_info=false																																												
 HVAC System Heating Other Fuel	None																																														
 HVAC System Heating None	None																																														
+HVAC Sizing Show Debug Info	No	ResidentialHVACSizing	show_debug_info=false																																												
+HVAC Sizing Show Debug Info	Yes	ResidentialHVACSizing	show_debug_info=true																																												
 HVAC System Cooling	None																																														
 HVAC System Cooling	"AC, SEER 8"	ResidentialHVACCentralAirConditionerSingleSpeed	seer=8.0	eer=7.3	shr=0.73	fan_power_rated=0.365	fan_power_installed=0.5	crankcase_capacity=0.0	crankcase_temp=55	eer_capacity_derate_1ton=1.0	eer_capacity_derate_2ton=1.0	eer_capacity_derate_3ton=1.0	eer_capacity_derate_4ton=1.0	eer_capacity_derate_5ton=1.0	capacity=autosize	dse=NA																															
-		ResidentialAirflow
```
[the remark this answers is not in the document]
joseph-robertson: @shorowit I removed them because they don't assign any argument values. Perhaps I don't total understand how the ApplyUpgrade measure works. Can you explain?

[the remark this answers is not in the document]
shorowit: With the original options_lookup file, if you select this option as an upgrade, the ResidentialAirflow measure will get called a second time after the ResidentialHVACCentralAirConditionerSingleSpeed measure is applied. Since no arguments are provided, it will use the same arguments as were specified for the creation of the existing building.

Imagine the upgrade was from a ducted HVAC system to a non-ducted HVAC system -- calling the ResidentialAirflow measure a second time will remove the ducts.

resources/options_lookup.tsv
```diff
@@ -1223,389 +1223,389 @@ HVAC System Is Shared	Only Heating Shared
...
+		ResidentialHVACSizing																																													
 HVAC System Shared Natural Gas	None																																														
 HVAC System Shared Natural Gas	Fan Coil Heating And Cooling	ResidentialHVACCentralSystemFanCoil	fan_coil_heating=true	central_boiler_fuel_type=gas																																											
 		ResidentialAirflow	has_hvac_flue=true																																												
-		ResidentialHVACSizing	show_debug_info=false																																												
+		ResidentialHVACSizing																																													
 HVAC System Shared Natural Gas	Fan Coil Cooling Only	ResidentialHVACCentralSystemFanCoil	fan_coil_heating=false	central_boiler_fuel_type=gas																																											
-		ResidentialAirflow	has_hvac_flue=true																																												
-		ResidentialHVACSizing	show_debug_info=false																																												
+		ResidentialAirflow
```
shorowit: I'm totally confused here. 

1) Why is the `has_hvac_flue` argument value being changed for some options? Was it wrong before?
2) Why do some "HVAC System Shared" options have the `has_hvac_flue` argument defined while others have the argument removed? There seems like no consistency?

joseph-robertson: For all the Shared Electricity system options I changed `has_hvac_flue=true` to `has_hvac_flue=false` because these systems would never have a flue.

Then I removed `has_hvac_flue` from all of the Cooling Only shared system options because `has_hvac_flue` would be supplied by the non-shared heating system option.

See https://github.com/NREL/OpenStudio-BuildStock/commit/d2844b2d137123940b172df409ac2747d2e24e1d.

shorowit: I see. Thanks for the explanation.

PR #418, opened 2020-03-02
Chicago AC saturation

JLReyna:
Addresses #[issue number here].

## Pull Request Description

During data source comparison, it was discovered that we probably have cooling saturation too low for Chicago. AHS is probably a better source for cooling saturation that we should use going forward. In the meantime, I pulled AC type saturation from AHS (None, Room AC , Central), and adjusted the distributions for each of these within the HVAC System Cooling tsv, but still using the efficiency breakdowns from RECS. Only CR04 was adjusted for the moment (which includes Chicago), using East North Central as an approximation for CR04 (the difference is the state of Wisconsin). In the future, we should probably separate saturation from efficiency in separate TSVs so we can use AHS for saturation and RECS for efficiency. 

## Checklist

Not all may apply:

- ~~[ ] Unit tests have been added or updated~~
- ~~[ ] Documentation has been modified appropriately~~
- ~~[ ] Any new options are added to `project_testing`~~
- ~~[ ] `project_testing` runs without any failures~~
- ~~[ ] No unexpected regression test changes~~


ejhw:
Please update the changelog with a brief note like above. Otherwise looks good!

Issue #20, opened 2016-08-04
Consolidate H1 regions?


joseph-robertson:
@cchriste proposes that we consolidate H1C1 OR, H1C1 WA, H1C2, and H1C3 into a single H1 region. Decision pending.

cc @jrobert1 @ewilson

joseph-robertson:
One advantage of keeping OR and WA separate (at least for H1), is that it could capture differences related to code adoption. E.g., 
usmap

joseph-robertson:
I doubt 2009 vs 2012 codes would even be a blip in the RBSA data. Though it could suggest a general trend of WA adopting codes before OR.

Seems like it'd be better to look at the RBSA queried results and see if there are significant differences in, e.g., insulation levels between the two states.

joseph-robertson:
Also, remember that we are no longer required to have every parameter tied to the same regions anymore. For example, if the envelope varied between WA and OR, we could make insulation parameters dependent on those regions while leaving other parameters (e.g., HVAC) dependent on heating zones.

joseph-robertson:
@ewilson, @shorowit,

Does this help in capturing differences between WA and OR?

Insulation Wall
image

joseph-robertson:
@jrobert1: Can you split the above table by heating zone so that we get tables specific to H1, H2, and H3? And do it for walls, attics, and windows? And commit text files instead of using images.

joseph-robertson:
@cchriste is going to revisit the mapping for wall insulation. Percentages of R-13 walls for pre-1950s homes seems large.

shorowit:
Completed.

PR #427, opened 2020-03-19
Update Infiltration.tsv based on the Residential Diagnostics Database.

afontani:
# Pull Request Description

**Companion PR:** [resstock-estimation #20](https://github.com/NREL/resstock-estimation/pull/20)

The purpose of this pull request is to diversify and update the infiltration housing characteristic.  Currently, infiltration is based on the [2014 House Simulation Protocols](https://www.nrel.gov/docs/fy14osti/60988.pdf). These protocols use a regression equation for infiltration. As a result, the probability distributions specified in the Infiltration.tsv are binary (only 0 and 1). In reality, the distributions are more continuous. 

In this pull request, the Infiltration.tsv has been updated with data from the [Residential Diagnostics Database](http://resdb.lbl.gov/) (ResDB). The cumulative distribution functions (CDFs) for air change at 50 Pa (ACH50) have been scraped from their website. Each CDF from the website was fit with a lognormal distribution. The fitted lognormal distribution is then used to assign probabilities into the infiltration bins in ResStock.

## Assumptions
It is worth pointing out a few of the assumptions in the distributions.

- The CDFs are for Single-Family Detached homes only. However, ResStock currently does not distinguish infiltration rates between building types.
- Data from certain IECC climate zones were copied to other climate zones. IECC climate zone 2A was copied to 1A. IECC climate zone 6A was copied to 7A. IECC climate zone 6B was copied to 7B.
- The ResStock vintage bins are more refined than the bins in ResDB. As a result certain decade bins had to be copied to other bins. CFDs from the 1950s were copied to the 1940s and the <1940. CFDs from the 2000s were copied to the 2010s.
- Homes are assumed to not be Weatherization Assistance Program (WAP) qualified and not ENERGY STAR certified.
- When obtaining the CDFs, the floor area value was set to be the `total_ffa` and `unit_ffa` for the Geometry Floor Area characteristic in options_lookup.tsv.

## Checklist

Not all may apply:


afontani:
After scraping the data from ResDB, the figure attached have some quality metrics on the goodness of fit for the lognormal distributions.
image

afontani:
## Dependency Selection
ResDB has a lot of different values that could be a dependency in the Infiltration.tsv file.

- Geometry Building Type RECS
- Vintage
- Geometry Foundation Type
- Geometry Floor Area
- Geometry Stories
- IECC Climate Zone 

Using each of these as a dependency would result in a file of approximately 100k rows.  To reduce the number of rows (and thus the number of dependencies), an analysis was performed to quantify their contribution towards modifying the probability distributions. For each dependency, the L2 distance was used to quantify the amount the dependency changes the distribution. The following algorithm was used.

1. Calculate the average distribution for the USA using the bins in the Infiltration.tsv.
2. For each dependency candidate
    a. Add a dependency 
    b. Quantify how much the distribution changed when adding the dependency.
   c. Remove the dependency
3. Select the dependency that changes the average distribution the largest.
4. Repeat steps 2-3 until there are no other dependencies.

This process gives a ranked list of the dependencies and their importance to probabilities in the housing characteristic. Below is a table of the distances (larger is better) for the current housing characteristic in master and the new housing characteristic based on ResDB. 
image

The values from the current version in master are larger because of the binary nature of the numbers. The dependencies chosen for the new tsv are Vintage, Geometry Floor Area, and ASHRAE IECC Climate Zone 2004. The parameters that were not selected, their PDFs are averaged. This selection reduces the number of rows to ~12,000.





-

afontani:
## Impact on Infiltration rates

For the US, the median infiltration rate will increase.
**Master: ** 8.84 ACH50
**ResDB:** 13.07 ACH50

Below is a CDF plot of what is in ResStock now (Master) and the update (ResDB)
image

Issue #434, opened 2020-04-03
Buildings incorrectly missing HVAC systems


JLReyna:
During troubleshooting an LA100 issue, we discovered an issue where buildings that have been assigned `Option=Electricity` in `Heating Fuel.tsv` are being assigned `Option=None` in `HVAC System Heating Electricity.tsv` even though the dependencies in specify that `Dependency=HVAC System Is Heat Pump` is No and `Dependency=HVAC System Is Shared` is `Neither Shared` or `Only Cooling Shared`. Essentially, this means our fraction of buildings without heating systems does not match `Heating Fuel.tsv`, and we are underestimating heating, in some regions (e.g. CR10) potentially up to 20%. This could also be an issue with other fuels, like natural gas, but this needs further investigation to confirm.

Recommendations:
- Completely restructure HVAC TSVs to simplify and ensure that distributions are accurate to source data
- Remove or simplify the dependencies that link distribution of electric heating technologies to cooling saturation
- Add in `resstock_probability` column as standard in all TSVs so that these issues can be identified and diagnosed more quickly
- Compare annual heating totals before and after fix to understand impact of missing heating on old projects.

shorowit:
Presumably this occurs because the source data has no samples for the combination of e.g., (heating fuel, climate zone, vintage), right? 

In the past, we had discussions about smoothing probabilities when the sampling is highly granular (@ejhw probably remembers this approach). Alternatively probabilities could filled in using probabilities from nearby combinations that do have samples.

joseph-robertson:
I don't think this is a tsv structure issue. If you look at `HVAC System Heating Electricity.tsv`, specifically rows where `Location Region=CR10`, `Vintage=1950s`, `Heating Fuel=Electricity`, `HVAC System Is Heat Pump=No`, and `HVAC System Is Shared=Neither Shared`, you see that the option is 100% `None`.

JLReyna:
@joseph-robertson and that's a problem. If we've said `Heating Fuel = Electricity` and not `Heating Fuel =None`, it needs an electric heating technology assigned, but it is not getting any equipment because `Option=None` is 100%  in `HVAC System Heating Electricity.tsv` and it's not getting a shared system or a heat pump. On the row you called out, the `resstock_probability` is 0.1%, so that fraction of the housing stock is not getting electric heating equipment even though it's included in the `Heating Fuel.tsv`

joseph-robertson:
[...]

PR #442, opened 2020-04-27
Release 2_2_4 patch

joseph-robertson:
## Pull Request Description

Argument `has_hvac_flue` for ResidentialAirflow wasn't being assigned when no heating system was sampled.

The proposed solution here is to move the `has_hvac_flue` argument from the ResidentialAirflow measure out into the HVAC equipment measures (furnace, boiler, unit heater, shared systems). The value assigned to `has_hvac_flue` will be set as an additional property on the Building model object, and then the airflow measure will parse it from the Building model object. If it doesn't exist, then the value will be set to `false`.

## Checklist

Not all may apply:


aspeake:
In addition to `Heating Fuel ==` `None` and `Other Fuel`, I am seeing no HVAC systems for homes with all other heating fuels when the following parameters are assigned:
- `HVAC System Is Heat Pump == Yes `
- `HVAC System Is Shared == Neither`

There is also a small fraction of homes with `Heating Fuel == Electricity` and `HVAC System Is Heat Pump == No` that do not get an HVAC system assigned

JLReyna:
@aspeake this is a mistake in the logic of the distributions (see #434 ). When we fix that issue so that homes assigned heating fuels are appropriately assigned heating systems, the proposed logic of the `has_hvac_flue` should work. I'm in the process of restructuring all of the HVAC TSVs, but it will probably be a month or so before I put in a PR. Are you wanting to see a quicker fix for the misassignment issue?

resources/measures/ResidentialHVACCentralSystemFanCoil/measure.rb
```diff
@@ -62,6 +69,9 @@ def run(model, runner, user_arguments)
 
     fan_coil_heating = runner.getBoolArgumentValue("fan_coil_heating", user_arguments)
     central_boiler_fuel_type = HelperMethods.eplus_fuel_map(runner.getStringArgumentValue("central_boiler_fuel_type", user_arguments))
+    if fan_coil_heating
```
aspeake, line 72: Is this check needed? It looks like this logic is built into options_lookup, at least for the one `ResidentialHVACCentralSystemFanCoil` option with `fan_coil_heating=false` (`Fan Coil Cooling Only`), the flue gets set to false.

resources/options_lookup.tsv
```diff
@@ -1222,275 +1222,275 @@ HVAC System Is Shared	Neither Shared
...
+		ResidentialAirflow																																													
 		ResidentialHVACSizing																																													
-HVAC System Heating Fuel Oil	Oil Boiler	ResidentialHVACBoiler	fuel_type=oil	"system_type=hot water, forced draft"	afue=0.76	oat_reset_enabled=false	oat_high=0.0	oat_low=0.0	oat_hwst_high=0.0	oat_hwst_low=0.0	design_temp=180.0	capacity=autosize	dse=NA																																		
-		ResidentialAirflow	has_hvac_flue=true																																												
+HVAC System Heating Fuel Oil	Oil Boiler	ResidentialHVACBoiler	fuel_type=oil	"system_type=hot water, forced draft"	afue=0.76	oat_reset_enabled=false	oat_high=0.0	oat_low=0.0	oat_hwst_high=0.0	oat_hwst_low=0.0	design_temp=180.0	capacity=autosize	dse=NA	has_hvac_flue=true																																	
+		ResidentialAirflow																																													
 		ResidentialHVACSizing																																													
 HVAC System Heating Wood	None																																														
 HVAC System Heating Wood	Wood Stove	ResidentialHVACUnitHeater	fuel_type=wood	efficiency=0.6	fan_power=0	airflow_rate=0	capacity=autosize
```
aspeake, line 1496: Move flue argument from airflow to `ResidentialHVACUnitHeater`?

aspeake: ...This is meant to reference the `HVAC System Heating Wood` option

resources/measures/ResidentialHVACCentralSystemFanCoil/measure.rb
joseph-robertson: My thought was that you could use the fan coil measure for cooling only. Then also have, e.g., the furnace measure for heating only. In such a case, you wouldn't want to overwrite the fan coil's `has_hvac_flue=false` with the furnace's `has_hvac_flue=true`.

resources/options_lookup.tsv
joseph-robertson: Oops, missed one. Good catch.

resources/measures/ResidentialHVACCentralSystemFanCoil/measure.rb
aspeake: Makes sense

shorowit:
For consistency, the `has_water_heater_flue` argument could be moved to the appropriate water heater measures.

PR #447, opened 2020-05-12
Dist column name fix

rajeee:
Removes the numeric value from the object name in the hot-water distribution measure.

[description here]

When running buildstockbatch simulation with the water volume output variable, currently it results in 1000s of columns because the column names are all unique for each building since the name includes a numeric value. This fixes the issue by removing the numeric value from the column name and making the column names the same for all buildings. 

Not all may apply:


resources/measures/ResidentialHotWaterFixtures/measure.rb
```diff
@@ -404,7 +404,7 @@ def remove_existing(runner, space, obj_names)
       found = false
       obj_names.each do |obj_name|
         next if not space_equipment.name.to_s.start_with? obj_name
-        next if space_equipment.name.to_s.include? "=" # TODO: Skip dummy distribution objects; can remove once we are using AdditionalProperties
+        next if space_equipment.name.to_s.include? "dist" # TODO: Skip dummy distribution objects; can remove once we are using AdditionalProperties
```
shorowit, line 407: How about this instead?
`next if space_equipment.additionalProperties.getFeatureAsDouble('dist_hw').is_initialized`

rajeee: Yeah, that looks much better! I didn't feel good about string check, but decided to follow suit anyway, haha!

Issue #444, opened 2020-05-07
Custom weather file integration and DST dates


whiphi92:
When using custom weather files from historical years, the DST dates are different from what is in optionslookup.tsv. So, if I am using AMY weather files I would need to change the DST entry of the Location object for every different year. Is there a workaround for this? 

I know I have been told not to change the weather filename entry of the Location object in optionslookup.tsv so I assumed it would be the same for other entries in the Location object.

joseph-robertson:
I think this should be a relatively easy fix/update. New to OS 3.0.0 are methods for returning daylight saving start/end dates from the epw file object. We can use these to override user-input daylight saving start/end dates for simulations that are run using AMY weather files.

shorowit:
There has been a lot of changes related to DST since this issue was created. Does the issue still exist?

ejhw:
Yes, still exists. This is about avoiding the need to manually change DST start/end dates for different AMY years in options_lookup.tsv.

ejhw:
Or maybe @joseph-robertson  implemented it in #448  ?

PR #448, opened 2020-05-12
Addresses #444, use daylight saving dates in epw header

joseph-robertson:
Addresses #444.

## Pull Request Description

Currently when simulating using AMY weather files, the daylight saving start/end dates that are used do not correspond to the actual year. This changes that so start/end dates from the epw header are read if they exist.

Note that AMY weather files will need to be updated to contain daylight saving start/end dates in their headers. See http://climate.onebuilding.org/papers/EnergyPlus_Weather_File_Format.pdf for reference.

## Checklist

Not all may apply:


resources/measures/HPXMLtoOpenStudio/resources/location.rb
```diff
@@ -119,6 +119,20 @@ def self.apply_dst(model, runner, dst_start_date, dst_end_date)
...
 
         dst = model.getRunPeriodControlDaylightSavingTime
+
+        if epw_file.daylightSavingStartDate.is_initialized and epw_file.daylightSavingEndDate.is_initialized # HOLIDAYS/DAYLIGHT SAVINGS row in epw file
+          dst_start_date = epw_file.daylightSavingStartDate.get
+          dst_end_date = epw_file.daylightSavingEndDate.get
+          dst_start_date_month = dst_start_date.monthOfYear
+          dst_start_date_day = dst_start_date.dayOfMonth
+          dst_end_date_month = dst_end_date.monthOfYear
+          dst_end_date_day = dst_end_date.dayOfMonth
```
shorowit, line 129: Why not simply use `model.getRunPeriod.setUseWeatherFileDaylightSavings(true)`?

shorowit: Actually, if you don't specify the above, it might [default to true](https://bigladdersoftware.com/epx/docs/9-3/input-output-reference/group-location-climate-weather-file-access.html#field-use-weather-file-holidays-and-special-days-1). So simply adding the DST dates to the EPW might be enough?

joseph-robertson: Good questions. I'll check.

afontani:
A general question is how does DST work for different weather files? I am just thinking of AZ, which does not follow DST.

shorowit:
Good point @afontani. And the house may be in AZ but the weather file is from a nearby state (or vice versa)... 😬

resources/measures/HPXMLtoOpenStudio/resources/location.rb
joseph-robertson: So, `model.getRunPeriod.getUseWeatherFileDaylightSavings` defaults to true. But then `model.getRunPeriodControlDaylightSavingTime` shows start/end dates different from those in the epw header. Not sure if this is a bug...

afontani:
Would a possible solution be to have an argument in the measure that is the observed DST (observe_dst=true/false)? This way we pass in the dates and then we tell the model to observe those dates.  This way in options_lookup.tsv we could assign the argument with the State.tsv. Every stat would have observe_dst=true except AZ (observe_dst=False).

joseph-robertson:
@whiphi92 had a good proposed solution (sorry it wasn't captured here): move the assignment of `dst_start_date` and `dst_end_date` from the Location tsv to the County tsv. See the changes to options lookup in this PR.

resources/measures/HPXMLtoOpenStudio/resources/location.rb
joseph-robertson: I found this note in the forward translator for RunPeriod:
```
  // ETH@20121219 - We also do not support pulling the daylight savings period in from the weather
  // file. Explicit RunPeriodControl:DaylightSavingTime objects already override this field, so
  // making this change only affects files without such objects (and likely doesn't change results
  // for those since weather files from the E+ website have no such designations).
```

joseph-robertson:
This will be captured in ResStock-HPXML: https://github.com/NREL/OpenStudio-BuildStock/tree/restructure-v3

Issue #44, opened 2017-03-01
Multifamily data collection


ejhw:
List of useful data for building out multifamily building capabilities:

- Steven Winter Associates "Boiler Survey Data Bonanza" http://homeenergy.org/show/article/nav//id/2094

ejhw:
@joseph-robertson the boiler data linked above may be relevant to your hydronic efforts

PR #470, opened 2020-07-23
Update plug load multipliers

afontani:
## Pull Request Description

The purpose of this pull request is to address the over approximation of plug loads in Multi-Family and Single-Family Attached dwelling units. We have seen the over approximation Multi-Family loads from AMI data and in RECS 2015 enduse national totals.
image

The [resstock-estimation](https://github.com/NREL/resstock-estimation) companion Pull Request [#44](https://github.com/NREL/resstock-estimation/pull/44)

### Plug Loads
New options have been added to account for different multipliers for RECS building types.

### Usage Level
The Option=Average was removed as it was a duplicate of the Option=Medium for the housing characteristics that depend on Usage Level.tsv and Option=Average is not used in the national or testing project.

As a result, several dependants of Usage Level.tsv. needed to be updated. The list of housing characteristics that are dependent on Usage Level.tsv is itemized below. The rows where Dependency=Usage Level being "Average" were removed.

- Clothes Dryer
- Cooking Range
- Dishwasher
-  Hot Water Fixtures
- Refrigerator

### option_lookup.tsv updates

**Plug Loads:** The word "Usage" was added to the naming of the Plug Load.tsv options. New options for plug loads are as follows:

_New Options:_
- 33% Usage
- 40% Usage
- 66% Usage
- 80% Usage
- 132% Usage
- 159% Usage 

**Usage Level:** The "Average" option was removed.

**Line Endings:** Some line endings were renormalized to keep a consistent format.

## Checklist

Not all may apply:


joseph-robertson:
So this change isn't preferred over this PR? https://github.com/NREL/resstock-estimation/pull/45

ejhw:
> So this change isn't preferred over this PR? [NREL/resstock-estimation#45](https://github.com/NREL/resstock-estimation/pull/45)

We may want to do both. I guess I'd propose (all national runs):

1. Run with these .66 and .80 multipliers
2. Run with the new SF and MF plug load equations. (with original 1.0 multipliers)
3. Consider calculating new multipliers based on results of #2.

@afontani can tell us if this is getting to be too many runs 😁

afontani:
@joseph-robertson @ejhw: I think we should do both. It looks like from the new equations the plug loads are a lot less sensitive to household members for multifamily.

This will probably be a bit of an iterative process. I think the path @ejhw suggested is the right one. In terms of computer allocation for the runs, this might be a bit too much. 

I would like to suggest we use the latest national run to calculate the difference in the plug loads based on the new equations. This should inform if a substantial change to the multipliers is desired.

ejhw:
> I would like to suggest we use the latest national run to calculate the difference in the plug loads based on the new equations. This should inform if a substantial change to the multipliers is desired.

That's fine with me.

PR #483, opened 2020-09-03
Water Heater GEB

joseph-robertson:
Addresses #[issue number here].

## Pull Request Description

Companion PR: https://github.com/NREL/resstock-estimation/pull/150

Adding in new GEB related features, including:

- Allowing tanks to use either a mixed or stratified tank model (additional user argument)
- Allowing setpoint to follow an hourly (8760) schedule rather than only fixed values.
- Allowing HPWH operating mode to be scheduled (either "HP Only" or "standard"). This allows the elements to be disabled during peak periods.
- Adding unmet shower (an unmet load metric for WHs) calculations into ResStock runs to quantify any unmet WH loads.

In addition to these GEB features, a few new HPWH options, corresponding the AO Smith's current product line, are added to options lookup as potential upgrade options.

## Checklist

Not all may apply:


ejhw:
@jmaguire1 Would you propose using stratified tank models by default in ResStock? What would the energy and runtime impacts of that decision be?

jmaguire1:
> @jmaguire1 Would you propose using stratified tank models by default in ResStock? What would the energy and runtime impacts of that decision be?

I think my recommendation would be that we probably don't make that change right away. Noel and I did a lot of work improving the run time of the stratified tank model, so it no longer increases the run time of the simulation by 2-3x. That's available in EnergyPlus 9.4 on. If ResStock was using E+ 9.4, I think the run time impact would be pretty minimal.

In terms of energy consumption, I'd expect there to be a small impact on the energy consumption, with a stratified tank model consuming slightly less. In my past analysis (which is only one location/draw profile), there was a 2% reduction in water heating energy consumption. I think this is largely due to the bottom few nodes of the tank below the lower element never get heated up to the setpoint. It'd also have some impact on how much hot water is delivered and potentially any unmet loads.

ejhw:
Sounds good, thanks @jmaguire1 . We'll be on E+ 9.4 once we migrate to OS-HPXML.

project_testing/housing_characteristics/Water Heater Efficiency.tsv
```diff
@@ -1,3 +1,3 @@
...
-0.071429	0.071429	0.071429	0.071429	0.071429	0.071429	0.071429	0.071429	0.071429	0.071429	0.071429	0.071429	0.071429	0.071429
-# Created by: sources\recs\2009\tsv_maker.py
\ No newline at end of file
+"Option=Electric Heat Pump, 80 gal"	Option=Electric Premium	Option=Electric Standard	Option=Electric Tankless	Option=FIXME Fuel Oil Indirect	Option=Fuel Oil Premium	Option=Fuel Oil Standard	Option=Natural Gas Premium	Option=Natural Gas Standard	Option=Natural Gas Tankless	Option=Other Fuel	Option=Propane Premium	Option=Propane Standard	Option=Propane Tankless	"Option=Electric Standard, Scheduled"	"Option=Natural Gas Standard, Scheduled"	"Option=Electric Heat Pump, 50 gal, Scheduled"	"Option=Electric Heat Pump, 80 gal, Scheduled"
+0.055556	0.055556	0.055556	0.055556	0.055556	0.055556	0.055556	0.055556	0.055556	0.055556	0.055556	0.055556	0.055556	0.055556	0.055556	0.055556	0.055556	0.055556
```
ejhw, line 2: Are these just being added for testing and not project_national?

resources/options_lookup.tsv
```diff
@@ -536,47 +536,59 @@ Water Heater Fuel	Fuel Oil
...
+Water Heater Control	WH Tankless Scheduled	ResidentialHotWaterHeaterTankless																																													
+Water Heater Control	WH Stratified Tank NEWE_Net_2020	ResidentialHotWaterHeaterTank	setpoint_type=scheduled	schedules_directory=./resources	tank_model_type=stratified	setpoint_schedule=DR_DHW_Sch_NEWE_Net_2020.csv																																									
+Water Heater Control	WH Tankless NEWE_Net_2020	ResidentialHotWaterHeaterTankless																																													
+Water Heater Control	WH HP NEWE_Net_2020	ResidentialHotWaterHeaterHeatPump	setpoint_type=scheduled	operating_mode_type=constant	schedules_directory=./resources	setpoint_schedule=DR_DHW_Sch_NEWE_Net_2020.csv	operating_mode_schedule=hourly_operating_mode_schedule.csv																																								
+Water Heater Control	None																																														
+Water Heater Efficiency	Electric Benchmark	ResidentialHotWaterHeaterTank	fuel_type=electric	tank_volume=auto	setpoint_temp=125	location=auto	capacity=auto	energy_factor=auto	recovery_efficiency=0.98	offcyc_power=0	oncyc_power=0	has_water_heater_flue=false
```
ejhw: Do we really need this many options? What is "NEWE_Net_2020"?

jmaguire1:
This is now, finally, ready for review! @ejhw @aspeake

project_testing/housing_characteristics/Water Heater Efficiency.tsv
jmaguire1: @ejhw All of the scheduled options for WH control are being added in so people could potentially look at them as upgrades, but national penetration is going to be set to zero. The intent with this branch is to add in a bunch of new functionality we could use, but not leverage any of it just yet until we have a better idea of national penetration of some of these things (if they are indeed even nonzero).

resources/options_lookup.tsv
jmaguire1: Some of these options were from Matt Leach's branch, I took a bunch of them out that were specific to his project before we finalized this.

measures/PowerOutage/measure.xml
```diff
@@ -1,9 +1,25 @@
...
+C:/OpenStudio-BuildStock-WH-GEB-ToMerge/resources/measures/HPXMLtoOpenStudio/resources/util.rb:1309: syntax error, unexpected tCONSTANT, expecting keyword_end
+...| water_heater.is_a? OpenStudio::Model::WaterHeaterStratifie...
+...                               ^
+C:/OpenStudio-BuildStock-WH-GEB-ToMerge/resources/measures/HPXMLtoOpenStudio/resources/util.rb:1410: syntax error, unexpected tCONSTANT, expecting keyword_end
+...| water_heater.is_a? OpenStudio::Model::WaterHeaterStratifie...
+...                               ^
+C:/OpenStudio-BuildStock-WH-GEB-ToMerge/resources/measures/HPXMLtoOpenStudio/resources/util.rb:1531: syntax error, unexpected tCONSTANT, expecting keyword_end
+...| water_heater.is_a? OpenStudio::Model::WaterHeaterStratifie...
+...                               ^
+C:/OpenStudio-BuildStock-WH-GEB-ToMerge/resources/measures/HPXMLtoOpenStudio/resources/util.rb:1921: syntax error, unexpected keyword_end, expecting end-of-input</error>
```
shorowit, line 18: 👆 Looks like something needs to be fixed in resources/measures/HPXMLtoOpenStudio/resources/util.rb.

aspeake: This should fix it: https://github.com/NREL/resstock/pull/483/commits/c2b4083b4f533ebdb453af3d60f25ba6e0b2f596

resources/options_lookup.tsv
```diff
@@ -687,53 +687,59 @@ Water Heater Fuel	Fuel Oil
...
 		ResidentialAirflow																																													
-Water Heater Efficiency	"Electric Heat Pump, 80 gal"	ResidentialHotWaterHeaterHeatPump	storage_tank_volume=80.0	setpoint_temp=125.0	location=auto	element_capacity=4.5	min_temp=45.0	max_temp=120.0	cap=0.979	cop=2.4	shr=0.98	airflow_rate=480.0	fan_power=0.178	parasitics=8.5	tank_ua=4.0	int_factor=1.0	temp_depress=0.0																														
+Water Heater Efficiency	"Electric Heat Pump, 80 gal"	ResidentialHotWaterHeaterHeatPump	storage_tank_volume=80.0	setpoint_temp=125.0	location=auto	element_capacity=4.5	min_temp=45.0	max_temp=120.0	cap=0.979	cop=2.4	shr=0.98	airflow_rate=480.0	fan_power=0.178	parasitics=8.5	tank_ua=4.0	int_factor=1.0	temp_depress=0.0	setpoint_type=constant	schedules_directory=./resources	setpoint_schedule=hourly_setpoint_schedule.csv	operating_mode=standard	operating_mode_type=constant	operating_mode_schedule=hourly_operating_mode_schedule.csv
```
ejhw, line 728: Can we change "operating_mode_type" to "operating_mode_schedule_type" ? Seems a little ambiguous otherwise.

ejhw (changes requested):
Thanks @jmaguire1 ! This looks good.

1. So HPWHs always use the stratified tank model? And non-HPWH tanks use mixed by default? 
2. When would you recommend using stratified for all tanks? What would the runtime implications be?

resources/options_lookup.tsv
jmaguire1: Yes! Just changed the name in the latest commit (https://github.com/NREL/resstock/commit/3369a83c1a7c63c5b80ab4171a41a8788d58c052) to avoid any confusion here.

HPWH will always be a stratified tank, yes. I haven't had any luck modeling them with a mixed tank because the HP performance is a function of the temperature in the bottom half of the tank (adjacent to the condenser), not the entire tank, and the control logic is so finnicky it would be really hard to approximate with a mixed tank. I didn't want this merge to change any of the energy consumption results so I left all of the other water heaters as mixed tanks, at least for now.

We actually could look at switching to stratified tanks now, but there'll be some energy impacts associated with it. @nmerket and I added in the new stratified tank formulation into E+ version 9.2, so there should be pretty negligible run time impacts from making the switch now. There will however be impacts to energy consumption and thermal comfort (this branch also adds in unmet shower metrics) associated with making the switch. I was planning to try to get this merged in with mixed tanks so there's no energy impact and then start looking at the impact of making the switch to stratified tanks by default.

ejhw: Thanks @jmaguire1 . That plan sounds good. I do like the idea of having stratified tanks by default so users don't need to remember to do that when applying WH DF measures.

ejhw (approved):

shorowit:
@jmaguire1 What was the cause for the small differences in energy?

resources/measures/ResidentialHotWaterHeaterTank/measure.rb
```diff
@@ -166,6 +204,39 @@ def run(model, runner, user_arguments)
...
+      # run_period_end = Time.new(assumed_year, run_period.getEndMonth, run_period.getEndDayOfMonth, 24)
+      # sim_hours = (run_period_end - run_period_start) / 3600
+
+      # year_description = model.getYearDescription
+      # n_days = Constants.NumDaysInYear(year_description.isLeapYear)
+      # if (sp_hourly_array.length != sim_hours) and (sp_hourly_array.length != n_days * 24)
+      #  runner.registerError("Hourly water heater setpoint schedule length must equal to a full year")
+      #  return false
+      # end
+      # return true
```
shorowit, line 232: Why is all of this commented out?

resources/measures/ResidentialHotWaterHeaterTankless/tests/ResidentialWaterHeaterTankless_Test.rb
```diff
@@ -75,36 +75,54 @@ def test_retrofit_replace
...
   end
 
+  def test_retrofit_replace_hpwh_scheduled_sp_with_tankless_gas
+    args_hash = {}
+    args_hash['fuel_type'] = Constants.FuelTypeGas
+    expected_num_del_objects = { 'WaterHeaterStratified' => 1, 'ScheduleFixedInterval' => 3, 'ScheduleConstant' => 4, 'WaterHeaterHeatPumpWrappedCondenser' => 1, 'CoilWaterHeatingAirToWaterHeatPumpWrapped' => 1, 'FanOnOff' => 1, 'OtherEquipment' => 2, 'OtherEquipmentDefinition' => 2, 'EnergyManagementSystemSensor' => 10, 'EnergyManagementSystemActuator' => 7, 'EnergyManagementSystemTrendVariable' => 3, 'EnergyManagementSystemProgram' => 2, 'EnergyManagementSystemProgramCallingManager' => 1 }
+    expected_num_new_objects = { 'WaterHeaterMixed' => 1, 'ScheduleConstant' => 2 }
+    expected_values = { 'InputCapacity' => 29307107, 'ThermalEfficiency' => 0.754, 'Setpoint' => 125, 'OnCycle' => 7.38, 'OffCycle' => 7.38, 'FuelType' => Constants.FuelTypeGas }
+    _test_measure('SFD_2000sqft_2story_FB_GRG_UA_3Beds_2Baths_Denver_HPWH_scheduled_SP.osm', args_hash, expected_num_del_objects, expected_num_new_objects, expected_values, 1)
```
shorowit, line 117: This test file, and a couple others, weren't committed to the repo.

resources/measures/ResidentialHotWaterHeaterTankless/tests/ResidentialWaterHeaterTankless_Test.rb
```diff
@@ -75,36 +75,54 @@ def test_retrofit_replace
...
-    expected_num_new_objects = { 'WaterHeaterMixed' => 1, 'ScheduleConstant' => 1 }
+    expected_num_del_objects = { 'WaterHeaterStratified' => 1, 'ScheduleConstant' => 6, 'CoilWaterHeatingAirToWaterHeatPumpWrapped' => 1, 'FanOnOff' => 1, 'WaterHeaterHeatPumpWrappedCondenser' => 1, 'OtherEquipment' => 2, 'OtherEquipmentDefinition' => 2, 'EnergyManagementSystemProgramCallingManager' => 1, 'EnergyManagementSystemProgram' => 2, 'EnergyManagementSystemActuator' => 7, 'EnergyManagementSystemSensor' => 10, 'EnergyManagementSystemTrendVariable' => 3 }
+    expected_num_new_objects = { 'WaterHeaterMixed' => 1, 'ScheduleConstant' => 2 }
     expected_values = { 'InputCapacity' => 29307107, 'ThermalEfficiency' => 0.754, 'Setpoint' => 125, 'OnCycle' => 7.38, 'OffCycle' => 7.38, 'FuelType' => Constants.FuelTypeGas }
     _test_measure('SFD_2000sqft_2story_FB_GRG_UA_3Beds_2Baths_Denver_HPWH.osm', args_hash, expected_num_del_objects, expected_num_new_objects, expected_values, 1)
   end
 
+  def test_retrofit_replace_hpwh_scheduled_sp_with_tankless_gas
```
shorowit, line 111: When you run this test (and several others), you get this OpenStudio warning: 
`[utilities.idf.WorkspaceObject] <2> Attempt to write a disconnected WorkspaceObject out to Idf.` 

This may indicate the code is not operating properly.

resources/measures/ResidentialHotWaterHeaterTankless/tests/ResidentialWaterHeaterTankless_Test.rb
shorowit: I was mistaken, these files were indeed committed.

resources/measures/HPXMLtoOpenStudio/resources/waterheater.rb
```diff
@@ -1016,7 +1218,7 @@ def self.get_plant_loop_from_string(model, runner, plantloop_s, unit)
     end
 
     model.getPlantLoops.each do |plant_loop|
-      next if plant_loop.name.to_s != plantloop_s
+      next if pl.name.to_s != plantloop_s
```
joseph-robertson, line 1221: @afontani Here's the issue.

resources/measures/ResidentialHotWaterHeaterTank/measure.rb
jmaguire1: @shorowit : I was going to add some additional error checking, got a little stuck getting it working, and left it commented out. I'm actually thinking I'll start a new branch to fix this, then get working on porting all of this code into HPXML.

jmaguire1:
> @jmaguire1 What was the cause for the small differences in energy?

@shorowit: There's a slight change in the WH setpoint as part of this. What we've been doing since we added deadband control the to the WH is artificially increasing the setpoint slightly (by deadband/2), with the goal of maintaining consistency with what we had in BEopt where the WH always perfectly maintained setpoint in terms of annual energy consumption. But it's not what actually happens with real WHs and it's confusing for a user to be specifying a setpoint schedule and seeing in the results that they're getting a slightly higher setpoint, so I made a change so that the WH setpoint is just the WH setpoint, with no increase. This leads to the slight decrease in WH energy consumption and some small second order impacts on HVAC due to changes in tank losses.

ejhw:
@shorowit Where can I see the diff in energy? Is it this file https://github.com/NREL/resstock/pull/483/commits/b7c7d41aaa01235eba3ad918c5e984a6b9a33754 ?

@jmaguire1 I'm not seeing the setpoint change in options_lookup. Is that happening behind the scenes? Is it just for stratified tank or all WHs?

jmaguire1:
> @shorowit Where can I see the diff in energy? Is it this file [b7c7d41](https://github.com/NREL/resstock/commit/b7c7d41aaa01235eba3ad918c5e984a6b9a33754) ?
> 
> @jmaguire1 I'm not seeing the setpoint change in options_lookup. Is that happening behind the scenes? Is it just for stratified tank or all WHs?

I've been looking at https://github.com/NREL/resstock/actions/runs/1046573826 , in the comparisons zip file there's a results_output.csv. That has the delta between develop and this branch in there, and the results are consistent with what I'd expect (a slight decrease in WH energy consumption based on the slightly lower sepoint temperature)

ejhw:
@jmaguire1 Thanks, so if I'm looking at this right (dividing by base), it's ~1% decrease across those samples for both elec and gas WHs.

jmaguire1:
@ejhw Yup that's correct, it's a pretty small change. But it is a change and one that might not have been obvious to everyone as this pull request is primarily about adding new functionality that isn't installed in the baseline, which I think was why @shorowit asked why there's any change at all.

shorowit:
I think the setpoint code change is appropriate -- for a given setpoint value, this is how a water heater behaves. However, I've always wondered if, when we talk about a 125F or 130F number, that really represents A) the water heater setpoint or B) an average measured temperature from the field. If the latter, then we would really want to modify our assumed setpoint value in conjunction with the code change and the results would be unchanged. @jmaguire1 Any idea here?

jmaguire1:
@shorowit: It's much more A, but when we're using a mixed tank you're fully averaging the tank temperature which has some impact on the energy consumption compared to real tanks. Average tank temperature in the field/lab is always lower than the setpoint since there's a 10 F single sided deadband for most water heaters (so if you're set to 125 F, temperature by the sensor drops to 115 F before the heater kicks on, then it reheats to 125 F)

One of the things added in this PR was adding in the functionality to use stratified tank models, but I think with this E+ version we're not fully prepared to leverage the runtime updates Noel and I worked on. So the plan going forward when we port this to HPXML would be to switch to stratified tank models as the default, at least for ResStock use cases, which will have some additional energy consumption impacts.

shorowit:
Thanks @jmaguire1. FYI, we are quite close to [updating ResStock to the latest EnergyPlus version](https://github.com/NREL/resstock/pull/604).

PR #416, opened 2020-02-26
Geospatial

afontani:
This pull request addresses #415.

## Pull Request Description

This pull request is aimed at adding some spatial specification of ResStock samples. Currently, there are only two spatial descriptions in the housing characteristics: 1) Location (the epw weather file regions), and 2) Location Region (RECS defined Custom Regions) This pull request creates 10 housing characteristic files that help spatially define the location of a sampled housing unit. 

### New or updated spatial housing characteristics
Data for these characteristics are based on the American Community Survey (ACS) 5-year 2016 tables B25001 and B25004. Unit counts and vacant unit counts are specified at the census tract level. The distributions are based on the occupied unit counts (table B25001 minus table B25004).

1. ASHRAE IECC Climate Zone 2004.tsv
2. State.tsv
3. County.tsv
4. PUMA.tsv
5. Census Division.tsv
6. Census Region.tsv.tsv
7. Location Region.tsv
8. Building America Climate Zone.tsv
9. Location.tsv
10. ISO RTO Region.tsv

Based on these new spatial specifications, PUMS data is used to provide more granular distributions of certain housing characteristics.

### Housing Characteristics Updated with the Public Use Microdata Sample (PUMS)
The following housing characteristics have been updated using 2017 PUMS.

1. Geometry Building Type ACS
2. Geometry Building Type RECS
3. Heating Fuel
4. Vintage ACS
5. Vintage

The creation of all housing characteristics has been automated by `tsv_maker` classes. Supporting data has been uploaded to s3.

## Checklist

Not all may apply:

- [ ] Unit tests have been added or updated
- [x] All rake tasks have been run, and pass
- [ ] Documentation has been modified appropriately
- [x] Any new options are added to `project_testing`
- [x] `project_testing` runs without any failures
- [ ] No unexpected regression test changes
- [x] All tests are passing (green) on circleci
- [x] The [changelog](https://github.com/NREL/OpenStudio-BuildStock/blob/master/CHANGELOG.md) has been updated appropriately
- [x] This branch is up-to-date with master

For more information on how to perform these checklist items, see the documentation's [Advanced Tutorial](https://resstock.readthedocs.io/en/latest/advanced_tutorial/index.html).

Issue #73, opened 2017-08-29
Output variable naming convention


ejhw:
Proposed changes to [SimulationOutputReport/measure.rb](https://github.com/NREL/OpenStudio-BuildStock/blob/master/measures/SimulationOutputReport/measure.rb). There may be some differences between Res and Com (e.g., end uses) but hope we can unite on the basics.

### Proposed changes:
1. General format & principles: `<fuel>_<end use>_<units>`. Fuel can be [electricity, natural_gas, other_fuel, source_energy, co2e]. "Site" is implied. Source and co2e are always aggregations of fuel types (calculated in post-processing). For example:
   - "electricity_total_kwh" instead of "total_site_electricity_kwh"
   - "electricity_net_kwh" instead of "net_site_electricity_kwh"
2. "lighting_interior" and "lighting_exterior" instead of "interior_lighting" and "exterior_lighting"
3. For residential, rename "water_systems" to "domestic hot water"

### Questions:
- [x] Does `OpenStudio::convert(total_val, os_units, desired_units)` interpret `MBtu` as 10e6 or 10e3.
- [ ] Is "total" the best name for whole-building consumption or is it too vague? Other options are "building", "whole_building", "facility", "meter"... OS sqlFile uses "sqlFile.electricityTotalEndUses".
- [ ] @rHorsey "total_site_energy_mbtu" or "total_site_energy_kwh"?
- [ ] @asparke2, okay to use "water_heating" instead of "water_systems"?

Eventually we'd like to split out "other fuel" into propane and fuel oil as well.

_Side note: Can we modify the code so these strings don't have to be changed in multiple places in [SimulationOutputReport/measure.rb](https://github.com/NREL/OpenStudio-BuildStock/blob/master/measures/SimulationOutputReport/measure.rb)?_

cc @joseph-robertson @shorowit @rHorsey @asparke2

shorowit:
The answer to your first question is 10e6.

shorowit:
water_systems is the official end use name you get from EnergyPlus.

ejhw:
@shorowit, yes, but is that what we want to use? water_heating seems more appropriate to me (or domestic_water_heating / service_water_heating). water_systems could be interpreted as pumping, etc.

ejhw:
@afontani I think the consensus was to not change now, but maybe revisit in July. Is that right?

afontani:
outdated.

Issue #63, opened 2017-06-19
Data: regional saturation factors for large uncommon loads


ejhw:
Would be beneficial for load modeling, especially for pool pumps.

ejhw:
Also especially gas fireplace, gas grill, and gas lighting.

PR #492, opened 2020-10-02
Update Geometry Foundation Type to be Based on RECS

afontani:
## Pull Request Description

This pull request updates the `Geometry Foundation Type.tsv` from the "[Building Foundation Design Handbook](https://www.osti.gov/biblio/6980439-building-foundation-design-handbook)" published in 1988 to RECS 2009. The new TSV has dependencies on IECC Climate and Moisture regions and Vintage ACS.

**Companion PRs:**
- [resstock-estimation #63](https://github.com/NREL/resstock-estimation/pull/63)
- [resstock-estimation #73](https://github.com/NREL/resstock-estimation/pull/73)

### Previous Data Source Documentation

TSV previously based on ORNL's "[Building Foundation Design Handbook](https://www.osti.gov/biblio/6980439-building-foundation-design-handbook)" published in 1988. 

image

### New Data Source
The new data source for `Geometry Building Type.tsv` is RECS 2009. Since the spatial restructure (PR #416), this housing characteristic is a bit out of date because the locations were redefined. As a result, this pull request also fixes the spatial issue created by PR #416.

#### Dependencies:
- ASHRAE IECC Climate Zone 2004
- Geometry Building Type RECS
- Vintage ACS

#### Assumptions
- All mobile homes have Pier and Beam foundations.
- Multi-family cannot have Pier and Beam and Heated Basements.
- Single-Family Attached buildings cannot have Pier and Beam foundations.
- Foundation types are the same for each building type except for mobile homes and the applicable options.
- If no foundation type specified, then sample has Pier and Beam foundation.
- RECS has a quirk where MF 5+ units always have foundation type fields set to N/A, whereas MF2-4 unit fields are allowed to have foundation types. Because we need to assume a foundation type for ground-floor MF units, we use the lumped SFD+SFA+MF2-4 distributions for MF5+. 
- A non-trivial percentage of MF units have neither basement, nor slab, nor crawl, which we map to pier and beam (aka open crawlspace). This seems unrealistic, but could represent MF units that are above parking garages or retail space, so we assume it is reasonable to model them as pier and beam open crawlspaces.

### Updates to Geometry Garage.tsv
Geometry Garage.tsv is updated in the pull request. Now that pier and beam foundations can now be sampled as a foundation type, the `BuildExistingModel` measure does not allow for pier and beam foundations with garages. As a result, a Geometry Foundation Type dependency was added to the Geometry Garage.tsv. 

A Geometry Foundation Type dependency is added to the `project_testing/Geometry Garage.tsv`.

**Assumptions:**
- Pier and beam foundations cannot have an attached garage
- Due to low sample counts and the need to split out pier and beam foundations, the samples with Crawl, Heated basement, Slab, and Unheated basement foundation types are lumped together.

## Checklist

Not all may apply:


ejhw:
See assumptions I added above. Would be nice to add these to the tsv files as well.

> - RECS has a quirk where MF 5+ units always have foundation type fields set to N/A, whereas MF2-4 unit fields are allowed to have foundation types. Because we need to assume a foundation type for ground-floor MF units, we use the lumped SFD+SFA+MF2-4 distributions for MF5+. 
> - A non-trivial percentage of MF units have neither basement, nor slab, nor crawl, which we map to pier and beam (aka open crawlspace). This seems unrealistic, but could represent MF units that are above parking garages or retail space, so we assume it is reasonable to model them as pier and beam open crawlspaces.

ejhw (approved):
👍

afontani:
@ejhw: Not a problem. Addressed `3415bd2`.

shorowit:
"A non-trivial percentage of MF units have neither basement, nor slab, nor crawl, which we map to pier and beam (aka open crawlspace). This seems unrealistic, but could represent MF units that are above parking garages or retail space, so we assume it is reasonable to model them as pier and beam open crawlspaces."

MF units, or MF buildings? If the former, wouldn't this be more likely to represent MF units above other MF units?

With respect to units over parking garages, it might be a good idea to create a new foundation type option for this. You could map it to pier and beam for now, but with the eventual switchover to HPXML, there is an ["other non-freezing space"](https://openstudio-hpxml.readthedocs.io/en/latest/hpxml_to_openstudio.html#hpxml-enclosure) that could be used to represent units over parking garages. As indicated it is modeled as floating with outdoor temperature but with a minimum of 40 deg-F.

ejhw:
Thanks @shorowit , that's a good question. I had been interpreting the fields for the **_building_** containing the household's dwelling unit, but I think you are right that it is for the **_unit_** itself.

From the codebook:

Name | Description
-- | --
CELLAR | Basement in   housing unit
CRAWL | Housing   unit over a crawl space
CONCRETE | Housing   unit over a concrete slab

We don't know the location of the unit in the building, which makes it tough to use these fields for MF. Also slabs wouldn't necessarily be on the ground floor. 

Once we move to single-unit MF modeling it might become easier to use these fields, but at least for now we still need distributions for foundation types for MF2-4 and MF5+ buildings

Options:
1. Use SFD+SFA distributions for SFD, SFA, MF2-4, and MF5+, and disallow pier&beam for MF2-4 and MF5+
image

2. Use SFD+SFA distributions for SFD, SFA. Use MF2-4 distributions for MF2-4 and MF5+ and disallow pier&beam for MF2-4 and MF5+. MF2-4 doesn't really have enough samples (only 722) to support us using it alone though when we have 3 dependencies.
image

So I vote option 1 until we move to  single-unit MF modeling and implement unit-specific tsvs.

cc @aspeake for awareness

ejhw:
@afontani I implemented option 1 since the logic was fresh in my mind: https://github.com/NREL/resstock-estimation/pull/63

ejhw:
@shorowit  do you want a chance to review what I implemented (https://github.com/NREL/resstock-estimation/pull/63)? Are you good with Option 1 above? Thanks again for raising the question.

shorowit:
@ejhw No, I don't need to review it. I'll defer to your judgment on how you decided to address my comment.

PR #523, opened 2020-12-16
Fix  surface area reporting for collapsed buildings

aspeake:
## Pull Request Description

Fixes below grade wall and door area calculations in `SimulationOutputReport` when `minimal_collapsed = true` for MF and SFA buildings. Previously, `units_represented` was not used to scale foundation wall or door area, and therefore areas were underestimated when building geometries were collapsed.

## Checklist

Not all may apply:

- [x] The [changelog](https://github.com/NREL/resstock/blob/master/CHANGELOG.md) has been updated appropriately


measures/SimulationOutputReport/measure.rb
```diff
@@ -776,16 +776,6 @@ def get_cost_multiplier(cost_mult_type, model, runner)
 
             cost_mult += UnitConversions.convert(surface.grossArea, "m^2", "ft^2")
           end
-
-        elsif cost_mult_type == "Wall Area, Below-Grade (ft^2)"
```
joseph-robertson, line 780: This logic essentially got moved to down below in the second section? Could all cost multiplier blocks be moved to down below so there is only a single block for each one?

aspeake: I don't think so, the first block calculates cost multipliers for each unit then scales by `units_represented`, the second section calculates for spaces not belonging to a unit. So since the below grade spaces are not associated with units it's not necessary to include in the first block, but is for most other outputs.

joseph-robertson: I see. But what about below-grade wall surfaces adjacent to finished basements? Aren't those associated with units?

joseph-robertson: Maybe we need a SimOutputReport test for SFA with [finished basement](https://github.com/NREL/resstock/blob/develop/resources/measures/ResidentialGeometryCreateSingleFamilyAttached/measure.rb#L94)?

aspeake: Good point. I guess I just combined non-unit foundation walls (eg, corridors) and the unit-specific foundation walls into one section, which we could do for other outputs if we wanted.

aspeake: I agree on the unit test for SFA, I'll add one

joseph-robertson: So I'm confused why any tests with SFD finished basement are passing then?

aspeake: All below-grade wall area is summed here: https://github.com/NREL/resstock/blob/799f8e5d3ebb29dff7528f72233151a2768d4235/measures/SimulationOutputReport/measure.rb#L905 

This is what is returned for SFD and non-collapsed SFA and MF buildings. If `minimal_collapsed`, then it scales the collapsed walls here: https://github.com/NREL/resstock/blob/799f8e5d3ebb29dff7528f72233151a2768d4235/measures/SimulationOutputReport/measure.rb#L929

joseph-robertson: Ok, so you intentionally have no presence of "Wall Area, Below-Grade (ft^2)" in the section above where `units_represented` is used. I guess, to me, it just seems a little inconsistent. The section above is for calculations made for surfaces adjacent to units, whereas the section below is for surfaces adjacent to spaces not associated with units. With the exception of "Wall Area, Below-Grade (ft^2)". Seems like for consistency we should either maintain the two "sections" or have only one "section".

aspeake: Right, so it is a special case for below-grade wall area just because this approach requires iterating through all wall area (unit and non-unit specific) then scaling the unit-specific walls, so it doesn't really break apart easily like the other outputs. I can definitely re-work each output to keep it to one section, and add the `unit.spaces.each do |space|` loop to each output variable block in the second section, I think that would help with structure/readability.

However, these changes are irrelevant once `singleunit` branch is merged, where `SimulationOutputReport` measure is very similar what is currently on `develop`

joseph-robertson: Good point.

joseph-robertson (approved):

PR #532, opened 2021-01-11
Update duct related calcs in airflow.rb

joseph-robertson:
## Pull Request Description

Addresses https://trello.com/c/VmcGehGM/62-duct-leakage-total.

Based on ASHRAE Standard 152:

Change equations for determining fraction of ducts outside conditioned space:
```
  def self.get_location_frac_leakage(location_frac, stories)
    if location_frac == Constants.Auto
      if stories == 1
        location_frac_leakage = 1
      else
        location_frac_leakage = 0.75
      end
    else
      location_frac_leakage = location_frac.to_f
    end
    return location_frac_leakage
  end
```

Change equations for determining supply/return surface areas:
```
  def self.get_duct_supply_surface_area(ffa)
    return 0.27 * ffa # ft^2
  end
```
```
  def self.get_duct_return_surface_area(ffa, num_returns)
    if num_returns < 6
      b_r = 0.05 * num_returns
    else
      b_r = 0.25
    end
    return b_r * ffa # ft^2
  end
```

Multiply duct areas by the fraction of ducts outside conditioned space:
```
    supply_surface_area = Airflow.get_duct_supply_surface_area(unit_ffa) * ducts.supply_area_mult * f_out
    return_surface_area = Airflow.get_duct_return_surface_area(unit_ffa, ducts.num_returns) * ducts.return_area_mult * f_out
```

Then determine supply/return ua by:
```
      supply_ua = supply_surface_area / supply_r
      return_ua = return_surface_area / return_r
```
which is used in the ducts ems subroutine.

Finally, determine supply/return loss by:
```
      supply_loss = f_out * ( supply_leakage + ah_supply_leakage )
      return_loss = f_out * ( return_leakage + ah_return_leakage )
```
which is used by sizing.rb.

## Checklist

Not all may apply:

- [ ] Unit tests have been added or updated
- [ ] All rake tasks have been run, and pass
- [ ] Documentation has been modified appropriately
- [ ] Any new options are added to `project_testing`
- [ ] `project_testing` runs without any failures
- [ ] No unexpected regression test changes
- [ ] All tests are passing
- [ ] The [changelog](https://github.com/NREL/resstock/blob/master/CHANGELOG.md) has been updated appropriately
- [ ] This branch is up-to-date with develop

For more information on how to perform these checklist items, see the documentation's [Advanced Tutorial](https://resstock.readthedocs.io/en/latest/advanced_tutorial/index.html).

PR #536, opened 2021-01-21
ResStock-HPXML: Upgrade costs

joseph-robertson:
## Pull Request Description

Introduce new `UpgradeCosts` reporting measure, which is essentially the upgrade portion of the old `SimulationOutputReport` measure. We now retrieve cost multipliers directly from the `hpxml` instead of the `model`.

The `UpgradeCosts` measure:
* has no arguments
* gets cost multipliers from `in.xml`, the defaulted version of `existing.xml` (the exported hpxml from `BuildExistingModel`)
* multiplies cost value (saved by `ApplyUpgrade`) by cost multiplier value
* has unit tests that operate on test osw files corresponding to test files for the old `SimulationOutputReport` measure

## Checklist

Not all may apply:

- [ ] All tests are passing
- [ ] The [changelog](https://github.com/NREL/resstock/blob/master/CHANGELOG.md) has been updated appropriately


measures/UpgradeCosts/measure.rb
```diff
@@ -0,0 +1,282 @@
...
+    elsif cost_mult_type == 'Floor Area, Attic (ft^2)'
+      hpxml.frame_floors.each do |frame_floor|
+        next unless frame_floor.is_thermal_boundary
+        next unless frame_floor.is_interior
+
+        if frame_floor.is_floor
+          next unless [HPXML::LocationAtticVented, HPXML::LocationAtticUnvented].include?(frame_floor.interior_adjacent_to)
+        elsif frame_floor.is_ceiling
+          next unless [HPXML::LocationAtticVented, HPXML::LocationAtticUnvented].include?(frame_floor.exterior_adjacent_to)
+        end
```
shorowit, line 208: The attic floor is always considered a ceiling framefloor, so you should be able to simplify this as:

```
        next unless frame_floor.is_thermal_boundary
        next unless frame_floor.is_interior
        next unless frame_floor.is_ceiling
        next unless [HPXML::LocationAtticVented, HPXML::LocationAtticUnvented].include?(frame_floor.exterior_adjacent_to)
```

measures/UpgradeCosts/measure.rb
```diff
@@ -0,0 +1,282 @@
...
+      end
+    elsif cost_mult_type == 'Door Area (ft^2)'
+      hpxml.doors.each do |door|
+        cost_mult += door.area
+      end
+    elsif cost_mult_type == 'Duct Surface Area (ft^2)'
+      hpxml.hvac_distributions.each do |hvac_distribution|
+        hvac_distribution.ducts.each do |duct|
+          cost_mult += duct.duct_surface_area
+        end
```
shorowit, line 235: @ejhw was interested in only [including ducts outside conditioned space](https://github.com/NREL/resstock/pull/532#issuecomment-760284928). So I suggest A) excluding any ducts where the location is living space or conditioned basement, and B) renaming this to Unconditioned Duct Surface Area.

measures/UpgradeCosts/measure.rb
```diff
@@ -0,0 +1,282 @@
...
+      end
+    elsif cost_mult_type == 'Size, Heating System (kBtu/h)'
```
shorowit, line 237: Wouldn't this code end up including the heating supplemental system capacity? I don't see where you are accounting for that.

measures/UpgradeCosts/measure.rb
```diff
@@ -0,0 +1,282 @@
...
+      hpxml.heating_systems.each do |heating_system|
+        next if heating_system.fraction_heat_load_served == 1.0
+
+        heating_systems[heating_system] = heating_system.fraction_heat_load_served
+      end
+      if heating_systems.size == 2
+        heating_systems = heating_systems.sort_by{|k, v| v}
+        cost_mult += heating_systems.values[0] # lowest frac
+      end
+    elsif cost_mult_type == 'Size, Cooling System (kBtu/h)'
```
shorowit, line 259: I assume you know this, but this code won't actually work for ResStock because ResStock uses autosized HVAC systems and the calculated capacities are not currently populated in the in.xml HPXML file. Your tests are working only because you're referencing HPXML files with hardsized systems.

measures/UpgradeCosts/tests/upgrade_costs_test.rb
```diff
@@ -0,0 +1,123 @@
...
+      'Roof Area (ft^2)' => 1510.0,
+      'Window Area (ft^2)' => 108.0 * 2 + 72.0 * 2,
+      'Door Area (ft^2)' => 40.0 * 2,
+      'Duct Surface Area (ft^2)' => 150.0 + 50.0,
+      'Size, Heating System (kBtu/h)' => 64.0,
+      'Size, Heating Supplemental System (kBtu/h)' => 0.0,
+      'Size, Cooling System (kBtu/h)' => 48.0,
+      'Size, Water Heater (gal)' => 40.0,
+    }
+    _test_cost_multipliers('base.xml', cost_multipliers)
```
shorowit, line 33: I would feel _much_ more comfortable with this measure if we could use the [same test buildings](https://github.com/NREL/resstock/blob/develop/measures/SimulationOutputReport/tests/simulation_output_report_test.rb) as were used in the SimulationOutputReport. 
A) Those tests have far better coverage of different situations. 
B) We'd be testing _actual_ HPXML files that ResStock generates, so we'd see for example that the HVAC cost multipliers code won't currently work for ResStock. 
C) We should get the exact same values for most of the cost multipliers.

measures/BuildExistingModel/measure.rb
```diff
@@ -225,7 +225,7 @@ def run(model, runner, user_arguments)
       measures['BuildResidentialHPXML'][0][step_value.name] = value
     end
 
-    measures['HPXMLtoOpenStudio'] = [{ 'hpxml_path' => File.expand_path('../existing.xml') }]
+    measures['HPXMLtoOpenStudio'] = [{ 'hpxml_path' => File.expand_path('../existing.xml'), 'output_dir' => File.expand_path('..'), 'debug' => true }]
```
shorowit, line 228: We should change the HPXMLtoOpenStudio measure so that you don't have to set debug=true just to get the in.xml file, since that should have some runtime impact. Either we should A) always generate in.xml (this is my preference assuming the added runtime is really small) or B) create a new measure argument specifically asking if this file should be generated.

shorowit (commented):
This is a really good start. Overall the approach is what I expected to see. Some comments below.

It's also gratifying to see how much simpler the code is to calculate the cost multipliers using the HPXML! The previous code was ~550 LOC while your new code is ~100 LOC.

measures/UpgradeCosts/measure.rb
joseph-robertson: I know that. The FIXMEs were used to indicate that. Where can we get calculated system sizes? Can we not populate them in the in.xml HPXML file?

measures/UpgradeCosts/tests/upgrade_costs_test.rb
joseph-robertson: So we'd generate HPXML versions of those test files? Any ideas where that should happen?

measures/UpgradeCosts/measure.rb
joseph-robertson: Good point. I added `next if heating_system.fraction_heat_load_served != 1.0`.

measures/UpgradeCosts/tests/upgrade_costs_test.rb
shorowit: Yes. Would it make sense to add OSWs to this tests dir and generate the HPXMLs on the fly for each test?

joseph-robertson: That seems like a good idea.

measures/UpgradeCosts/measure.rb
shorowit: You could technically have a single heating system/heat pump that serves less than 100%, but maybe ResStock never generates such an HPXML.

ejhw: We may want that ability in the future for zonal electric baseboard.

measures/UpgradeCosts/measure.rb
shorowit: That is the goal, but it's a bit complicated. I'll give some thought to this. I don't think this issue needs to hold up the PR, but we'll obviously need to address it eventually.

measures/UpgradeCosts/measure.rb
joseph-robertson: I changed this to check for number of heating systems. If there are exactly 1, then it is the primary heating system (and there is no secondary heating system). If there are exactly 2, then the heating system with the least load fraction served is the secondary heating system.

PR #628, opened 2021-06-30
ResStock-HPXML: Comparison enhancements

joseph-robertson:
Resolves #[issue number here].

## Pull Request Description

[description here]

## Checklist

Not all may apply:

- [ ] All tests are passing


resources/hpxml-measures/workflow/tests/compare.py
```diff
@@ -57,7 +56,51 @@ def results(self, base_folder, feature_folder, export_folder, groupby=[]):
         df = df.astype(int)
 
       df = df.fillna('NA')
-      df.to_csv(os.path.join(export_folder, file))
+      df.to_csv(os.path.join(export_folder, file))    
+
+      # Get results charactersistics of groupby columns
+      if file == 'results_characteristics.csv':
+        if 'build_existing_model.geometry_building_type_recs' not in groupby:
```
joseph-robertson, line 63: @aspeake Are you able to make this generic so that you could aggregate by any column in results_characteristics.csv (not just geometry_building_type_recs)?

resources/hpxml-measures/workflow/tests/compare.py
```diff
@@ -57,7 +56,51 @@ def results(self, base_folder, feature_folder, export_folder, groupby=[]):
...
+                            'diff':'n/a',
+                            '% diff':'n/a'},
+                            index=['simulation_count'])
+    sims_df[groupby] = 'n/a'
+    deltas = pd.concat([sims_df, deltas])
+    for group in groupby:
+      first_col = deltas.pop(group)
+      deltas.insert(0, group, first_col)
+
+    deltas.to_csv(os.path.join(export_folder, 'aggregate_results.csv'))
```
joseph-robertson, line 103: @aspeake Can you update this filename to be the original filename, then underscore, then the column you are grouping by? (See below where this is happening for plot filenames.)

resources/hpxml-measures/workflow/tests/compare.py
aspeake: @joseph-robertson it is currently set up to always group by RECS building type, but also allows any additional groupby fields to be added alongside building type. I figured it makes sense to default building type, but can make that require an input as well, thoughts?

resources/hpxml-measures/workflow/tests/compare.py
aspeake: @joseph-robertson yes, we still need to write it with some new prefix (ie `aggregate_results_*`) to differentiate from the results_output.csv though, right?

joseph-robertson: @aspeake I don't think so. These results are exported to a separate folder (i.e., `comparisons` instead of `base`).

aspeake: @joseph-robertson But it looks like there is already a `comparisons/results_output.csv` file that gets generated, which compares buildings 1:1

resources/hpxml-measures/workflow/tests/compare.py
joseph-robertson: I'd suggest limiting the number of groupby columns to 1. I wouldn't default to any groupbys here since this script is stored on OS-HPXML.

joseph-robertson:
Improvements:
* For the 1:1 plot, just use the one broken out by building type
* Include another plot of groupby sum by building type

shorowit:
> 
> 
> Improvements:
> 
>     * For the 1:1 plot, just use the one broken out by building type
> 
>     * Include another plot of groupby sum by building type

For the latter, we discussed summing _for each location_, so each graph would have 6 (or however many weather files we downselect to on the CI) points instead of 1.

PR #443, opened 2020-04-30
ResStock-HPXML

joseph-robertson:
## Pull Request Description

Subtree to: https://github.com/NREL/OpenStudio-HPXML/pull/372.

Closes #774.

## Checklist

Not all may apply:

- [ ] Unit tests have been added or updated
- [ ] All rake tasks have been run, and pass
- [ ] Documentation has been modified appropriately
- [ ] Any new options are added to `project_testing`
- [ ] `project_testing` runs without any failures
- [ ] No unexpected regression test changes
- [ ] All tests are passing (green) on circleci
- [ ] The [changelog](https://github.com/NREL/OpenStudio-BuildStock/blob/master/CHANGELOG.md) has been updated appropriately
- [ ] This branch is up-to-date with master

For more information on how to perform these checklist items, see the documentation's [Advanced Tutorial](https://resstock.readthedocs.io/en/latest/advanced_tutorial/index.html).

PR #710, opened 2021-09-22
Fixed Supplemental Heat capacity affecting Design Heat Load

### LINKED ISSUE #709: Fixed supplemental heat capacity applies to heat pump as well
I want to autosize ASHPs for max load using a fixed capacity of supplemental heat to simulate cold climate central heat pumps without supplemental heat use. Currently, setting the `supplemental_capacity=0.0` in `options_lookup.tsv` causes the overall design heat load to be set to 0 [here](https://github.com/NREL/resstock/blob/develop/resources/measures/HPXMLtoOpenStudio/resources/hvac_sizing.rb#:~:text=if%20not%20hvac.FixedSuppHeatingCapacity.nil%3F). This means that the heat load will actually be set to the (lower) cooling capacity [here](https://github.com/NREL/resstock/blob/develop/resources/measures/HPXMLtoOpenStudio/resources/hvac_sizing.rb#:~:text=heatCap_Rated%20%3C%20unit_final.Cool_Capacity), leading to a severely undersized heat pump.

I suggest something like moving `self.process_fixed_equipment()` to after` self.process_finalize()` so that fixed sizes on supplemental heating do not affect the design loads for autosizing other equipment.

zlee406:
Resolves #709 .

Previously, setting `supplemental_capacity` to a fixed value in `options_lookup.tsv` also changed the design heat load for the heat pump sizing, leading to unexpected heat pump sizing behavior.  Now, the supplemental heat capacity is set to a fixed value if `hvac.FixedSuppHeatingCapacity` is specified. Otherwise, it is set as `unit_final.Heat_Load` as before.

## Checklist

Not all may apply:

- [X] All rake tasks have been run, and pass
- [X] `project_testing` runs without any failures
- [ ] All tests are passing
- [X] This branch is up-to-date with develop

### DIFF
--- a/resources/measures/HPXMLtoOpenStudio/resources/hvac_sizing.rb
+++ b/resources/measures/HPXMLtoOpenStudio/resources/hvac_sizing.rb
@@ -1846,9 +1846,12 @@ class HVACSizing
       unit_final.Cool_Airflow = unit_final.Cool_Airflow * unit_final.Cool_Capacity / prev_capacity
     end
     if not hvac.FixedSuppHeatingCapacity.nil?
-      unit_final.Heat_Load = UnitConversions.convert(hvac.FixedSuppHeatingCapacity, 'ton', 'Btu/hr')
-    elsif not hvac.FixedHeatingCapacity.nil?
-      unit_final.Heat_Load = UnitConversions.convert(hvac.FixedHeatingCapacity, 'ton', 'Btu/hr')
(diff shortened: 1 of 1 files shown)


joseph-robertson:
Thanks for the contribution, but after review we've decided to close this because:
1. In general, fixed capacities use is not recommended in ResStock, and
2. This code will become obsolete once [ResStock-HPXML](https://github.com/NREL/resstock/pull/443) is merged.

PR #780, opened 2021-12-03
ResStock-HPXML: Heat pump related fixes

joseph-robertson:
## Pull Request Description

Checklist:
- [x] Add `heat_pump_backup_heating_switchover_temp=0.0` in `options_lookup.tsv` for non dual fuel
- [x] Add `heat_pump_backup_heating_switchover_temp=30.0` in `options_lookup.tsv` for dual fuel
- [x] Include `electricity_heating_supplemental_kwh` in `column_mapping.csv`

## Checklist

Not all may apply:

- [ ] All tests are passing


shorowit:
I'm a bit confused as to why ResStock uses a switchover temperature for conventional HPs. I would think you only use a switchover heat pump for a DFHP, where there are economic gains to switching to gas at lower temperatures. For HPs with electric backup, the only possible reason to use a switchover temperature is if the COP drops below 1, which I assume happens far below 0 degrees. If the COP is above 1, then you might as well use the heat pump for as much output as you can get, and then supplement it with the electric backup, which is the default behavior when a switchover temperature is not provided. cc @ejhw

ejhw:
> I'm a bit confused as to why ResStock uses a switchover temperature for conventional HPs. I would think you only use a switchover heat pump for a DFHP, where there are economic gains to switching to gas at lower temperatures. For HPs with electric backup, the only possible reason to use a switchover temperature is if the COP drops below 1, which I assume happens far below 0 degrees. If the COP is above 1, then you might as well use the heat pump for as much output as you can get, and then supplement it with the electric backup, which is the default behavior when a switchover temperature is not provided. cc @ejhw

@shorowit I'm pretty sure it was inherited from BEopt :) 

I agree with your point and would be supportive of removing it from ResStock. However, I often hear anecdotal evidence that many conventional HPs are installed with a switchover temp around 30F, but I could be getting confusing with economic balance point settings for dual fuel systems.

shorowit:
@ejhw I always thought the ~30F was the _balance point_, i.e., the temperature at which the heat pump capacity cannot meet the home's load and backup heating is required. I didn't think it had anything to do with the _switchover temperature_, in which the heat pump is disabled. But maybe older HPs were so bad that their COP really did drop below 1 at these kinds of temperatures, I'm not sure offhand.

aspeake (commented):
Looks good. 

Another HP difference that I am seeing can have a significant influence: the crankcase temp and capacity differ for [develop](https://github.com/NREL/resstock/blob/develop/resources/measures/HPXMLtoOpenStudio/measure.rb#L2315-L2316) and [restructure](https://github.com/NREL/resstock/blob/restructure-v3/resources/hpxml-measures/HPXMLtoOpenStudio/resources/hvac.rb#L3506-L3508). Should this be corrected on develop or here? @ejhw @shorowit?

ejhw:
@aspeake I think `restructure` has a good reference with RESNET Publication No. 002-2017 so I'd go with that.

joseph-robertson:
@aspeake @ejhw @shorowit See https://github.com/NREL/resstock/pull/787 for crankcase temp/capacity changes.

Issue #731, opened 2021-10-25
Changes to make after switching to ResStock-HPXML


joseph-robertson:
Checklist:
- [x] [Andrew] Remove `test/column_mapping.csv` file and reference in `test/compare.py` (perhaps convert it to a template so we can support future output name changes?)
- [x] [Joe] Remove `python test/compare.py -m feature` from `config.yml`
- [x] [Joe] `project_testing/Schedules.tsv` to equal distribution of `Default` and `Stochastic` https://github.com/NREL/resstock-estimation/pull/187
- [x] [Joe] `project_testing/HVAC System Is Faulted.tsv` to equal distribution of `No` and `Yes` https://github.com/NREL/resstock-estimation/pull/192
- [x] [Joe] Add finished attic options back into `project_national` (and update `options_lookup.tsv` to actually model finished attics?) https://github.com/NREL/resstock/issues/820
- [x] [Joe] Enable rim joists (set this by default, and r-value to foundation wall r-value); we could potentially enable rim joists between floors by adjusting average ceiling height, but we won't for now
- [x] [Andrew] Remove tsv that combines wall type and exterior finish https://github.com/NREL/resstock-estimation/pull/190
- [ ] ~Automate creation of tsv files with commented out method calls; low priority; wait for RECS 2020 updates~
- [x] Clean up README.md:
  - [x]  "... arguments into OpenStudio Measures"
  - [x] two references for OpenStudio Measures
  - [x] "ongoing" EULP project
- [x] [Joe] Change timeseries in `base_results` back to kWh, kBtu (from MBtu) https://github.com/NREL/resstock/pull/841

shorowit:
There are a bunch of option names that should be changed after the switch too. The old names from develop are still being used on restructure for comparison purposes, but some of them are wrong or misleading. Some examples:
- [ ] Natural Ventilation: "Year-Round, 3 days/wk" => "Cooling Season, 7 days/wk"
- [ ] Hot Water Distribution: There is no differentiation between Copper & PEX or HomeRun vs TrunkBranch. Remove some options and rename all of them.

There may be others, I'm just documenting the ones I came across recently.

aspeake:
`ceiling_assembly_r` and `floor_over_garage_assembly_r` are non-zero for the `None` options of Insulation Ceiling and Insulation Floor to avoid hitting errors with finished attics and homes with floors above garages. These should both be set to 0, but some changes will be needed to do so:

[...]

PR #792, opened 2021-12-29
ResStock-HPXML: Update Assembly R-Values

aspeake:
## Pull Request Description
Separates out the R-value of exterior finishes and adds it to the rest of the wall assembly R-value. Previously, vinyl siding was assumed for all walls (R=0.6), but this was not updated when new finish options were introduced. A new argument `exterior_finish_r` was added to `ResStockArguments`, and all the `wall_assembly_r` values were reduced by the assumed R-value of vinyl siding. Also updates the assembly r values for frame floors to be more accurate.

`Shingle, Composition, Medium` and `Shingle, Asbestos, Medium` did not have existing R-values, and so the assumed R-value of 0.6 was left unchanged.

`Brick, 12-in, *` options were updated to use `CMU, 12-in Hollow` inputs as the base wall instead of  `CMU, 6-in Hollow`

## Checklist

Not all may apply:

- [ ] All tests are passing


test/base_results/upgrades/timeseries/results_output.csv
```diff
@@ -1727,9 +1727,9 @@ project_national,0.0,0.000511821,0.0020131625999999995,0.0,0.0,0.001228370399999
...
-project_national,0.0,6.824279999999999e-05,0.00020472839999999997,0.0,0.0,0.0,0.0,0.0004776996,0.001023642,0.0,0.003992203799999999,0.0,0.0,0.0014672201999999998,0.0,0.0077,0.0,0.0015695843999999999,0.0073019795999999994,0.0077,0.0,0.0,0.0
-project_national,0.0,6.824279999999999e-05,0.000170607,0.0,0.0,0.0,0.0,0.0004776996,0.001023642,0.0,0.003992203799999999,0.0,0.0,0.0014672201999999998,0.0,0.0,0.0,0.001535463,0.007199615399999999,0.0,0.0,0.0,0.0
+project_national,0.0,6.824279999999999e-05,0.00020472839999999997,0.0,0.0,0.0,0.0,0.0004776996,0.001023642,0.0,0.003992203799999999,0.0,0.0,0.0014672201999999998,0.0,0.0,0.0,0.001535463,0.0073019795999999994,0.0,0.0,0.0,0.0
+project_national,0.0,6.824279999999999e-05,0.000170607,0.0,0.0,0.0,0.0,0.0004776996,0.001023642,0.0,0.003992203799999999,0.0,0.0,0.0014672201999999998,0.0,0.0077,0.0,0.001535463,0.007267858199999999,0.0077,0.0,0.0,0.0
+project_national,0.0,6.824279999999999e-05,0.00013648559999999998,0.0,0.0,0.0,0.0,0.0004776996,0.001023642,0.0,0.003992203799999999,0.0,0.0,0.0014672201999999998,0.0,0.0,0.0,0.001535463,0.007199615399999999,0.0,0.0,0.0,0.0
```
shorowit, line 1732: @joseph-robertson Looks like we're getting CI diffs because the order is changing; can we prevent that? (And how about using fewer decimal places for these outputs?)

joseph-robertson: @aspeake I wonder if this is a DST issue. (I'll look into the decimal places issue.)

resources/options_lookup.tsv
```diff
@@ -412,65 +412,65 @@ Insulation Slab	"4ft R5 Under, Horizontal"	ResStockArguments	slab_perimeter_insu
...
-Insulation Floor	None	ResStockArguments	floor_over_foundation_assembly_r=3.1	floor_over_garage_assembly_r=3.1																								
-Insulation Floor	Uninsulated	ResStockArguments	floor_over_foundation_assembly_r=3.1	floor_over_garage_assembly_r=3.1																								
-Insulation Floor	Ceiling R-13	ResStockArguments	floor_over_foundation_assembly_r=15.6	floor_over_garage_assembly_r=15.6																								
-Insulation Floor	Ceiling R-19	ResStockArguments	floor_over_foundation_assembly_r=20.4	floor_over_garage_assembly_r=20.4																								
-Insulation Floor	Ceiling R-30	ResStockArguments	floor_over_foundation_assembly_r=28.1	floor_over_garage_assembly_r=28.1																								
-Insulation Floor	Ceiling R-38	ResStockArguments	floor_over_foundation_assembly_r=32.9	floor_over_garage_assembly_r=32.9																								
+Insulation Floor	None	ResStockArguments	floor_over_foundation_assembly_r=5.3	floor_over_garage_assembly_r=5.3
```
shorowit, line 415: @aspeake I realize you didn't start this approach, but I would feel more comfortable if we used zeroes for assembly R-values for None options to ensure these values are not accidentally used. This comment applies not just to floors but other surface types with None options. If these assembly R-values _are_ accidentally used for modeling, an assembly R-value of zero will trigger an error and we can then fix the underlying issue. @joseph-robertson, is there any reason we aren't doing this?

joseph-robertson: I can't think of why they'd need to be set to "Uninsulated" values instead of zero.

There would be a "None" option change for:
* `Insulation Ceiling`
* `Insulation Floor`
* others?

@shorowit @aspeake

shorowit: Those are the only two that I see.

aspeake: 👍 I'll update

shorowit:
@aspeake You can't remove the arguments from the None options, they are required measure arguments. That's why they were there to begin with.

In your previous commit, the CI gives the error message:

`Measure BuildExistingModel reported an error with [\"/__w/resstock/resstock/workflow/testing_upgrades/run0/run/existing.xml: Expected AssemblyEffectiveRValue to be greater than 0 [context: /HPXML/Building/BuildingDetails/Enclosure/FrameFloors/FrameFloor/Insulation, id: \\\"FrameFloor2Insulation\\\"]\"]`

As I said [earlier](https://github.com/NREL/resstock/pull/792#discussion_r791263299), "If these assembly R-values are accidentally used for modeling, an assembly R-value of zero will trigger an error and we can then fix the underlying issue. " This is the exact error I'm talking about. Either the assembly R-value of zero is being used when it shouldn't, or there is a legitimate reason that we are using an R > 0 that I am forgetting. You need to look at the building that triggers this error and understand what's going on.

aspeake:
> @aspeake You can't remove the arguments from the None options, they are required measure arguments. That's why they were there to begin with.
> 
> In your previous commit, the CI gives the error message:
> 
> `Measure BuildExistingModel reported an error with [\"/__w/resstock/resstock/workflow/testing_upgrades/run0/run/existing.xml: Expected AssemblyEffectiveRValue to be greater than 0 [context: /HPXML/Building/BuildingDetails/Enclosure/FrameFloors/FrameFloor/Insulation, id: \\\"FrameFloor2Insulation\\\"]\"]`
> 
> As I said [earlier](https://github.com/NREL/resstock/pull/792#discussion_r791263299), "If these assembly R-values _are_ accidentally used for modeling, an assembly R-value of zero will trigger an error and we can then fix the underlying issue. " This is the exact error I'm talking about. Either the assembly R-value of zero is being used when it shouldn't, or there is a legitimate reason that we need to use R > 0 that I am forgetting. You need to look at the building that triggers this error and understand what's going on.

@shorowit That error was being hit for buildings with slabs which get a `None` insulation floor option. Although `floor_over_foundation_assembly_r` is ignored for slabs, it apparently cannot be zero, so I think the insulation floor option for slabs needs to be changed to `Uninsulated`. I think there is a similar issue with finished/cathedral ceilings having a `None` insulation ceiling option.

shorowit:
> @shorowit That error was being hit for buildings with slabs which get a `None` insulation floor option. Although `floor_over_foundation_assembly_r` is ignored for slabs, it apparently cannot be zero, so I think the insulation floor option for slabs needs to be changed to `Uninsulated`. I think there is a similar issue with finished/cathedral ceilings having a `None` insulation ceiling option.

The HPXML file is creating a HPXML `FrameFloor` with an assembly R-value of zero. What floor surface does this `FrameFloor` represent in the slab-on-grade home? It sounds like you need to look closer.

aspeake:
> > @shorowit That error was being hit for buildings with slabs which get a `None` insulation floor option. Although `floor_over_foundation_assembly_r` is ignored for slabs, it apparently cannot be zero, so I think the insulation floor option for slabs needs to be changed to `Uninsulated`. I think there is a similar issue with finished/cathedral ceilings having a `None` insulation ceiling option.
> 
> The HPXML file is creating a `FrameFloor` with an assembly R-value of zero. What floor surface does this `FrameFloor` represent in the slab-on-grade home? It sounds like you need to look closer.

Decided to revert `ceiling_assembly_r` and `floor_over_garage_assembly_r` to non-zero to avoid errors with finished attics and floors over garages. Added a comment here to address the underlying problems: https://github.com/NREL/resstock/issues/731#issuecomment-1028125151

joseph-robertson (approved):

PR #854, opened 2022-03-17
Requesting user-defined output variables

joseph-robertson:
## Pull Request Description

Subtree'd from: https://github.com/NREL/OpenStudio-HPXML/pull/1003

## Checklist

Not all may apply:

- [x] No unexpected integration/regression test changes


.github/workflows/config.yml
```diff
@@ -80,7 +80,7 @@ jobs:
 
       - name: Run buildstockbatch
         run: |
-          pip install git+https://github.com/NREL/buildstockbatch.git@develop
+          pip install git+https://github.com/NREL/buildstockbatch.git@output-variables
```
shorowit, line 83: Need to change this back before merging.

project_testing/testing_baseline.yml
```diff
@@ -25,6 +25,7 @@ workflow_generator:
       simulation_control_run_period_end_month: 12
       simulation_control_run_period_end_day_of_month: 31
       simulation_control_run_period_calendar_year: 2007
+      debug: true
```
shorowit, line 28: Why?

project_testing/testing_baseline.yml
```diff
@@ -42,6 +43,8 @@ workflow_generator:
       include_timeseries_end_use_consumptions: true
       include_timeseries_emissions: true
       include_timeseries_total_loads: false
+      output_variables:
+        - name: Zone Mean Air Temperature
```
shorowit, line 47: That's a weird choice since you can get these temperatures using the built-in `include_timeseries_temperatures`.

project_testing/testing_baseline.yml
joseph-robertson: I switched this on so that I could look at rdd files for good output variable candidates.

project_testing/testing_baseline.yml
joseph-robertson: I previously had it set to "Zone People Occupant Count" but it was showing up as all zeros for some simulations. Any suggestions? (Also, with consideration to file size.)

project_testing/testing_baseline.yml
joseph-robertson: It's also consistent with telling `server_directory_cleanup` to retain all files.

project_testing/testing_baseline.yml
shorowit: That's weird. Occupants.tsv suggests that every home should have at least one occupant. Can you investigate why that is?

shorowit (approved):

Issue #200, opened 2019-01-18
Adam/Eric seeing "could not load openstudio" at end of integrity checks


joseph-robertson:

shorowit:
Yes, it requires the OpenStudio Ruby bindings as mentioned at:
https://resstock.readthedocs.io/en/latest/advanced_tutorial/installer_setup.html#windows-setup

joseph-robertson:
Does this prevent the integrity checks from actually working?

joseph-robertson:
And would there be a way to remove the "Wrote output file buildstock.csv" line?

shorowit:
Some of the integrity checks are not run if you get this error.

shorowit:
The sampling script writes that message. The integrity checks call the sampling script, which generates this message and writes the specified file, but then the integrity checks clean up after it and deletes the file.

PR #877, opened 2022-04-05
Reduce TSV Sizes

afontani:
## Pull Request Description

**Companion PR:** resstock-estimation [#200](https://github.com/NREL/resstock-estimation/pull/200)

As housing characteristics get bigger, we are moving to a compact writing style. The 6-digit float format requirement was put in place when most of the characteristics were not scripted. Now that there is a standard workflow for creating the characteristics, the formatting requirements are being relaxed. This change should allow for more accurate characteristic distributions because the exponential format can be used. The change should also stop round-off errors.

The change is mainly from `%.6f` to `%g` in `pandas.to_csv(...)` in the resstock-estimation repository.

Some CI tests were removed in the integrity checks that related to formatting checks in the characteristics.

## Housing Characteristic size reduction

**Project National**
- Before Change: 103MB
- After Change: 46MB

**Project Testing**
- Before Change: 648KB
- After Change: 560KB

## Checklist

Not all may apply:

~~[ ] `openstudio tasks.rb update_measures` has been run~~
- [ ] No unexpected integration/regression test changes


test/integrity_checks.rb
```diff
@@ -333,31 +333,7 @@ def check_parameter_file_format(tsvpath, n_deps, name)
...
-            raise "ERROR: Incorrect non float found in '#{name}', line '#{i}'."
-          end
-          # If float precision is not 6 digits, raise error
-          if float_precision != 6
-            raise "ERROR: Incorrect float precision found in '#{name}', line '#{i}'."
-          end
-        end
-      end
+      next
     else
```
shorowit, line 337: There is a small bug here, where by using `next`, the `i += 1` below will not get executed.

Suggest changing this from:
```
    if line.include? "\r\n"
      next
    else
```
to
```
    if not line.include? "\r\n"
```

shorowit (approved):
Looks good to me. Nice job updating all related tests, documentation, etc. Just one minor bug to address.

test/integrity_checks.rb
afontani: Thanks @shorowit!

test/integrity_checks.rb
```diff
@@ -332,33 +332,7 @@ def check_parameter_file_format(tsvpath, n_deps, name)
...
-      if i > 1
-        # Check float format
-        # Remove endline character and split the string into array
-        line = line.split("\r\n")[0].split("\t")
-        # For each non dependency entry check format
-        for j in n_deps..line.length() - 1 do
-          # Check for scientific format
-          if (line[j].include?('e-') || line[j].include?('e+') ||
-              line[j].include?('E-') || line[j].include?('E+'))
-            raise "ERROR: Scientific notation found in '#{name}', line '#{i}'."
```
joseph-robertson, line 346: Should we remove this check?

afontani:
> Do you have the numbers on how much this reduces total file size?

@joseph-robertson: Good suggestion. I updated the PR description. Project national reduced by about 50%.

test/integrity_checks.rb
afontani: The `%g` format allows for scientific notation. I actually think this is a good thing as it handles smaller values really well and helps with truncation error.

test/integrity_checks.rb
```diff
@@ -332,33 +332,7 @@ def check_parameter_file_format(tsvpath, n_deps, name)
...
-          end
-
-          begin # Try to get the float precision
-            float_precision = line[j].split('.')[1].length()
-          rescue NoMethodError
-            # Catch non floats
-            raise "ERROR: Incorrect non float found in '#{name}', line '#{i}'."
```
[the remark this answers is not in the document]
afontani: @joseph-robertson: I had to think about this for a second, but in the new format integers are acceptable. An example is our mapping characteristics (ex: State is dependant on County and results in a bunch of 1s and 0s). 

There is still a sum to 1.0 and positivity check, which should check for nans and odd values.

joseph-robertson (approved):

PR #639, opened 2021-07-15
ResStock-HPXML: Occupancy adjustments

shorowit:
Resolves #[issue number here].

## Pull Request Description

- [ ] **Appliances** - Incorporating # occupants adjustment factors on top of the base ANSI 301 equations. Should provide better agreement (but will never give perfect agreement).
- [ ] **MELs/MGLs/LULs** - The BAHSP-based equations are currently preserved in ResStock-HPXML, yielding perfect agreement. But should we switch to # occupants adjustment factors on top of the ANSI 301 equations as we're doing for appliances?
- [x] ~~**Lighting** - Equations have never been based on # occupants. ANSI 301 equations were [backported to current ResStock](https://github.com/NREL/resstock/pull/619), yielding perfect agreement.~~

## Checklist

Not all may apply:

- [ ] Unit tests have been added or updated
- [ ] All rake tasks have been run, and pass
- [ ] Documentation has been modified appropriately
- [ ] Any new options are added to `project_testing`
- [ ] `project_testing` runs without any failures
- [ ] No unexpected regression test changes
- [ ] All tests are passing
- [ ] The [changelog](https://github.com/NREL/resstock/blob/develop/CHANGELOG.md) has been updated appropriately
- [ ] This branch is up-to-date with develop

For more information on how to perform these checklist items, see the documentation's [Advanced Tutorial](https://resstock.readthedocs.io/en/latest/advanced_tutorial/index.html).

PR #915, opened 2022-05-06
Use OS-HPXML operational calculation

### LINKED ISSUE #851: Operational Calculation (occupancy-based loads)
Update `ResStockArguments` once https://github.com/NREL/OpenStudio-HPXML/issues/644 is resolved.

shorowit:
## Pull Request Description

https://github.com/NREL/OpenStudio-HPXML/pull/989

Closes #851. Our [previous occupancy adjustments](https://github.com/NREL/resstock/pull/639#issuecomment-881678901) resulted in clothes washer/dryer energy use that was ~30% lower than the Building America-based implementation. I realized that there was a better way to apply the adjustments to the 301-based implementation. This PR brings ResStock-HPXML and the original ResStock results into better alignment.

This PR only changes appliance and hot water usage. MELs/MGLs and LULs are identical to before.

## Checklist

Not all may apply:


measures/ResStockArguments/measure.rb
```diff
@@ -379,6 +379,9 @@ def run(model, runner, user_arguments)
 
     args_to_delete = args.keys - arg_names # these are the extra ones added in the arguments section
 
+    # Set operational calculation
+    args['occupancy_calculation_type'] = HPXML::OccupancyCalculationTypeOperational
```
shorowit, line 383: @joseph-robertson I don't know that there's a compelling use case for exposing this in options_lookup.tsv, but you're welcome to do so if you wish.

joseph-robertson: I'll leave that up to @ejhw.

joseph-robertson (approved):

measures/ResStockArguments/measure.rb
shorowit: If there is a use case for exposing this input, we can do so easily in a subsequent PR.

nmerket: It's been a while since we've built an ECE run. It was on a much older version of ResStock. I could see this being potentially useful to have a switch for asset calculations. I say add it if it doesn't cause too much touble.

nmerket:
Haha. Should have read to the end of the PR. This is fine. If we need it we can expose the input later.

PR #935, opened 2022-05-18
Test OS-HPXML PR 1090

shorowit:
## Pull Request Description

Testing https://github.com/NREL/OpenStudio-HPXML/pull/1090.

## Checklist

Not all may apply:

- [x] ~Documentation has been updated~
- [x] ~Changelog has been updated~


measures/ResStockArguments/measure.rb
```diff
@@ -41,7 +41,12 @@ def arguments(model)
       # Following are arguments with the same name but different options
       next if arg.name == 'geometry_unit_cfa'
 
-      args << arg
+      # Convert optional arguments to string arguments that allow Constants.Auto for defaulting
+      if !arg.required
+        args << OpenStudio::Measure::OSArgument.makeStringArgument(arg.name, false)
+      else
+        args << arg
+      end
```
shorowit, line 49: This allows options_lookup.tsv to continue to support "auto" for optional arguments.

measures/ResStockArguments/measure.rb
```diff
@@ -740,7 +733,7 @@ def run(model, runner, user_arguments)
       rescue
       end
 
-      if args_to_delete.include? arg_name
+      if args_to_delete.include?(arg_name) || (arg_value == Constants.Auto)
```
shorowit, line 736: And then at the very end, if there is a value of Constants.Auto, we exclude it from being passed to the BuildResHPXML measure (i.e., we don't provide a value for the optional arguments, which means they get defaulted).

shorowit (commented):
Changes to options_lookup.tsv:
- Adds `heat_pump_sizing_methodology` argument
- Adds various "present" arguments (e.g., `dishwasher_present`)
- Moves arguments from the "Has PV" meta-parameter to the "PV System Size" parameter
- Moves arguments from the "Clothes Washer Presence" meta-parameter to the "Clothes Washer" parameter

measures/ResStockArguments/measure.rb
```diff
@@ -427,20 +421,20 @@ def run(model, runner, user_arguments)
     end
 
     # Num Occupants
-    if args['geometry_unit_num_occupants'] == Constants.Auto
+    if args['geometry_unit_num_occupants'].to_s == Constants.Auto
       args['geometry_unit_num_occupants'] = Geometry.get_occupancy_default_num(args['geometry_unit_num_bedrooms'])
     else
-      args['geometry_unit_num_occupants'] = Integer(args['geometry_unit_num_occupants'])
+      args['geometry_unit_num_occupants'] = Integer(args['geometry_unit_num_occupants'].to_s)
```
joseph-robertson, line 427: Why string before integer?

measures/ResStockArguments/measure.rb
```diff
@@ -706,7 +699,7 @@ def run(model, runner, user_arguments)
 
     # Rim Joist Assembly R-Value
     rim_joist_assembly_r = 0
-    if args['geometry_rim_joist_height'].get > 0
+    if Float(args['geometry_rim_joist_height'].to_s) > 0
```
joseph-robertson, line 702: Why don't we need to `.get` anymore?

measures/ResStockArguments/measure.rb
shorowit: It used to be a [required string argument](https://github.com/NREL/OpenStudio-HPXML/blob/ac525ddca5d5178e089bb1104e86edc59ddbde4e/BuildResidentialHPXML/measure.rb#L260). Now it's an [optional double argument](https://github.com/NREL/resstock/blob/d0cdc51e9f4c9dc91b3af5727804c097073edbf0/resources/hpxml-measures/BuildResidentialHPXML/measure.rb#L257) (which is converted to an optional string argument to support "auto"), so I use `.to_s` to convert from optional string to string. I could have done `.get` instead.

shorowit:
@joseph-robertson I pulled in the latest OS-HPXML and now it gives errors: "Extra argument 'debug' specified for measure 'BuildResidentialHPXML'." What is the appropriate fix?

joseph-robertson (approved):

shorowit:
@joseph-robertson Thanks for fixing.

PR #938, opened 2022-06-02
Update dependencies

joseph-robertson:
## Pull Request Description

Updates example YML files with `os_version` and `os_sha` fields. This allows the user to change the version of OpenStudio that buildstockbatch uses. (If not provided, they default to `BuildStockBatchBase.DEFAULT_OS_VERSION` and `BuildStockBatchBase.DEFAULT_OS_SHA`.) For example you are using an older version of ResStock that depends on OpenStudio v3.3.0, but you need to use the latest version of buildstockbatch for some bugfix/feature (note this is not always guaranteed to work since buildstockbatch is not always backward compatible).

Also, require a minimum version of buildstockbatch. This protects against, for example, trying to use ResStock-HPXML with buildstockbatch v0.21.

Companion PR: https://github.com/NREL/buildstockbatch/pull/294.

## Checklist

Not all may apply:


.github/workflows/config.yml
```diff
@@ -89,7 +89,7 @@ jobs:
 
       - name: Run buildstockbatch
         run: |
-          pip install git+https://github.com/NREL/buildstockbatch.git@develop
+          pip install git+https://github.com/NREL/buildstockbatch.git@bump-version
```
rajeee, line 92: Don't you want to revert it back to "develop" ?

joseph-robertson: Yes, I'll do that after the companion PR is merged.

docs/read_the_docs/source/conf.py
```diff
@@ -12,22 +12,21 @@
...
-author = metadata['__author__']
+project = u'ResStock'
+copyright = u'2019, NREL'
+author = u'NREL'
 
+# The short X.Y version
+version = u''
 # The full version, including alpha/beta/rc tags
-release = metadata['__resstock_version__']
+release = u''
```
rajeee, line 29: Why is the version and release blank in here?

joseph-robertson: Hmm, I was follow [this](https://github.com/NREL/OpenStudio-HPXML/blob/master/docs/source/conf.py) as a guide. I don't think it ends up getting used?

rajeee: Okay.

rajeee (approved):
All looks fine.

PR #940, opened 2022-06-10
Latest OS-HPXML

shorowit:
## Pull Request Description

Relevant changes:
- Switches Kiva foundation model timestep from 'Hourly' to 'Timestep'; small increase in runtime for sub-hourly simulations. 
- Fixes possible incorrect autosizing of heat pump *separate* backup systems with respect to duct loads. 
- Fixes handling non-integer number of occupants when using the stochastic occupancy schedule generator. 
- Fixes units for Peak Loads (kBtu/hr, not kBtu) in annual results file.

FYI @afontani  @whiphi92 @joseph-robertson @rajeee @ejhw 

## Checklist

Not all may apply:


measures/ResStockArguments/measure.rb
```diff
@@ -599,7 +599,9 @@ def run(model, runner, user_arguments)
       # Error check MF & SFA geometry
       if !has_rear_units && ((corridor_position == 'Double-Loaded Interior') || (corridor_position == 'Double Exterior'))
         runner.registerWarning("Specified incompatible corridor; setting corridor position to 'Single Exterior (Front)'.")
-        corridor_position = 'Single Exterior (Front)'
+        # FIXME: This doesn't actually do anything because corridor_position is not used later.
+        # Remove code? Or should the code be moved up?
+        # corridor_position = 'Single Exterior (Front)'
```
shorowit, line 604: @aspeake @joseph-robertson Any idea how this code should be fixed?

rajeee:
@shorowit @whiphi92  Does this fully address the issues we were seeing on the latest EUSS run? Can we expect to have the HVAC size for primary and backup system to be quite close going forward?

shorowit:
> @shorowit @whiphi92 Does this fully address the issues we were seeing on the latest EUSS run? Can we expect to have the HVAC size for primary and backup system to be quite close going forward?

This fixes an issue on the OS-HPXML side that will help with some of the runs, but many of the runs will still have issues related to how the upgrades are defined. @whiphi92 will have to explain.

measures/ResStockArguments/measure.rb
aspeake: @shorowit I think the simplest fix would be to move the block above this line: https://github.com/NREL/resstock/blob/d2c21d59832b14560ef8da7da2e3059ff7aea69c/measures/ResStockArguments/measure.rb#L586

This scenario would fall under the `else` in the logic above, and should only apply to MF.

test/test_run_analysis.rb
```diff
@@ -71,7 +71,6 @@ def test_errors_already_exists
     yml = ' -y test/tests_yml_files/yml_bad_value/testing_baseline.yml'
     @command += yml
 
-    cli_output = `#{@command}`
```
joseph-robertson, line 74: @shorowit This was intentionally the same line repeated. Is there a way to tell rubocop (or whatever we're using) to ignore this one?

shorowit: Aha, thanks, I can address it then. It didn't dawn on me that the failing test could be the result of a change here.

shorowit: To be clear, rubocop isn't complaining that the same line is repeated, it's complaining the the `cli_output` in the first line is unused (because it is overridden by the following line). So the first line just needs to be `#{@command}` and then rubocop will be happy and the CI test will be happy.

tasks.rb
```diff
@@ -47,7 +52,17 @@ def display_usage(command_list)
   # Apply rubocop
   cops = ['Layout',
```
joseph-robertson, line 53: Aligns with OS-HPXML?

joseph-robertson (approved):

tasks.rb
shorowit: Yep. Helps to catch potential bugs (like the `corridor_position` bugfix in this case).

PR #946, opened 2022-06-23
New "MHSP xxx, Ducted" options

### LINKED ISSUE #924: HVAC Heat Pump Ducted Argument Location
Similar to the proposals in this comment [here](https://github.com/NREL/OpenStudio-HPXML/issues/1011#issuecomment-1102910006a), could we add  the `heat_pump_is_ducted` argument from the `HVAC Heating Type` option to the `HVAC Heating Efficiency` option so that it can be found with the other respective heat pump system arguments?

cc: @ejhw @joseph-robertson

joseph-robertson:
## Pull Request Description

Closes #924.

Because there are no MSHP options in the national project baseline stock, I think it makes sense to:
* create new "MSHP xxx, Ducted" options (and leave existing "MSHP xxx" options as not ducted)
* move `heat_pump_is_ducted` argument assignment from `HVAC Heating Type` to `HVAC Heating Efficiency`

Project yml files will therefore no longer require you to additionally specify `HVAC Heating Type|Ducted Heat Pump` for MSHP options that are ducted (instead, just use the "MSHP xxx, Ducted" option suffix).

## Checklist

Not all may apply:


whiphi92 (commented):
All looks good to me mechanically. However, if we are adding the ducted heat pump to the upgrades yamls we need to add logic so it isn't applied to non-ducted homes, right?

whiphi92 (changes requested):
In that case, this looks good. I would add this block to the upgrades going into the default upgrades ymls:

```
        apply_logic:
          - HVAC Has Ducts|Yes
```

PR #947, opened 2022-06-23
Model wood fuel when sampling "Other Fuel" for WH and "Other" for HVAC

### LINKED ISSUE #945: "Other fuel" water heaters prevent laundry energy consumption
Water heaters with fuels other than the big 4 (NG, electricity, propane, fuel oil) are about 0.7% of the stock according to RECS 2009. RECS 2015 and RECS 2020 combine kerosene with fuel oil and show 0.2% and 0.3% respectively using something other than NG, electricity, propane, fuel oil, or kerosene. These are low-ish numbers but not completely negligible. 

We don't model these fuels (e.g. wood, direct solar thermal, kerosene). This means that upgrading these water heaters (e.g. to HPWHs) results in apparent increased total energy consumption. That is explainable and reasonable. 

However, it looks like we also don't model these water heaters, which has downstream impacts. 

@rajeee found that every home with an "other fuel" water heater has 0 energy consumption for clothes washers and clothes dryers. When we upgrade the water heaters to HPWHs, most of the homes show typical washer and dryer energy consumption. This creates an average energy increase for clothes dryers when upgrading all water heaters to a HPWH - a result that doesn't make a lot of sense. 

@afontani found that the inputs for Water Heater Efficiency in options_lookup, when Option=Other Fuel, the arguments do not specify a water heater type (water_heater_type=none) and the setpoint is 0 degrees. He speculates there is no water flow to the washer, which means it can't run. And then I guess this means the dryer can't run either.

I assume this is not intended behavior?

joseph-robertson:
## Pull Request Description

Closes #945. This PR updates:
* `Water Heater Efficiency|Other Fuel` to model a wood storage water heater instead of no water heater.
* `HVAC Heating Efficiency|Other` to model a wood wall/floor furnace instead of no heating system.

A few things to consider here:
* Should `water_heater_type=storage water heater`? Should `water_heater_fuel_type=wood`?
* What about `HVAC Heating Efficiency|Other` (i.e., `Heating Fuel|Other Fuel`)? Currently no heating system is modeled.

Parameter | Option | Argument 1 | Argument 2 | Notes
-- | -- | -- | -- | --
Water Heater Efficiency | Other Fuel | `water_heater_type=storage water heater` | `water_heater_fuel_type=wood` | Wood storage water heater[^1]
HVAC Heating Efficiency | Other | `heating_system_type=WallFurnace` | `heating_system_fuel=wood` | Wood wall/floor furnace[^2]
Misc Hot Tub Spa | Other Fuel | `hot_tub_present=false` | `hot_tub_heater_type=none` | Leave alone (no wood option).
Misc Pool Heater | Other Fuel | `pool_heater_type=none` | | Leave alone (no wood option).

[^1]: Same option as "Natural Gas Standard", except for the fuel type. Somewhat arbitrarily choosing "storage water heater" here.

[^2]: Same option as "Fuel Wall/Floor Furnace, 60% AFUE", except for the fuel type. When `HVAC Heating Efficiency=Other`=>`HVAC Heating Type=Non-Ducted Heating` for the national project. Appears that stove, portable/fixed heater, wall/floor furnace, and fireplace are all unit heaters modeled in an identical way. So, somewhat arbitrarily choosing "WallFurnace" here.

## Checklist

Not all may apply:

- [x] No unexpected regression test changes on CI (checked comparison artifacts) **We don't sample `Water Heater Efficiency=Other Fuel` in any of our 350 regression test samples.**


rajeee:
Consensus to map other fuel to wood.

rajeee:
Also consider following similar approach for "other fuel" used in heating.

shorowit (commented):
The changes look reasonable to me, though I'd be concerned that the ResStock outputs for wood could be easily misinterpreted. Should all the fuel types that aren't directly modeled in ResStock (wood, coal, ...?) be aggregated into "other" output columns?

ekpresent:
@shorowit I have never seen nonzero output values for wood or coal in ResStock. Is this a thing that has ever come up or is expected to come up in the future? I don't know at what point they become zero.

shorowit:
Ah, thanks for pointing that out @ekpresent. I saw a Heating Fuel=Wood in options_lookup.tsv, so I thought resstock was already modeling some buildings w/ wood, but now I see that it's not referenced by project_national.

shorowit (approved):

ekpresent:
@shorowit wood is included in the characteristics distributions for things like heating fuel, but we only represent the energy consumption from electricity, natural gas, fuel oil, and propane - though again, I don't really know what happens behind the scenes to create that outcome.

PR #962, opened 2022-07-11
Clean up options_lookup.tsv in github actions

rajeee:
## Pull Request Description

In an effort to clean up `resources/options_lookup.tsv`, this sets a standard for what the file should look like. The `.gitattributes` determines the line ending style. Then trailing whitespace (tabs inserted by Excel), and ordering of values are automatically updated using github actions. The hope is that diffs for new features are easier to look at, conflicts are easier to resolve, and characteristics/options are easier to find.

### Updates to `options_lookup.tsv`
- Normalize line ending for options lookup to CRLF (like the housing characteristics).
- Remove leading and trailing whitespace
- Sort based on parameter and option

## Checklist

Not all may apply:


afontani:
@rajeee: Can we also remove trailing whitespace? I am mostly concerned with the extra tabs that get added by Excel.

.github/workflows/config.yml
```diff
@@ -8,6 +8,18 @@ on:
   workflow_dispatch:
 
 jobs:
+  clean-files:
+    runs-on: ubuntu-latest
+    steps:
```
joseph-robertson, line 13: @afontani I think you need a checkout step to begin with...

joseph-robertson: I wonder if you should just have this happen in the `update-results` job. Otherwise, you'd need to upload as artifact, then download as artifact, etc...

afontani: @joseph-robertson and I decided to change options_lookup upfront and then download it and run the other tests with the formatted file. Then commit it during the `update-results` job.

afontani:
@joseph-robertson : The functionality seems to be working. I am a little confused why the `test/base_results` files are showing changes. The comparisons are also showing a scattered set of results between the base and the feature branch. I don't expect there to be changes. I checked out the most recent files from `develop` at [259f732](https://github.com/NREL/resstock/pull/962/commits/259f732490bf123084f2be704e924b5c8978923f) and they were replaced in the following "Latest Results" commit.

Do you have any thoughts? See attached screenshot of the most recent artifact from [259f732](https://github.com/NREL/resstock/pull/962/commits/259f732490bf123084f2be704e924b5c8978923f).

Screen Shot 2023-06-24 at 8 44 00 AM

joseph-robertson:
> @joseph-robertson : The functionality seems to be working. I am a little confused why the `test/base_results` files are showing changes. The comparisons are also showing a scattered set of results between the base and the feature branch. I don't expect there to be changes. I checked out the most recent result files from `develop` at [259f732](https://github.com/NREL/resstock/pull/962/commits/259f732490bf123084f2be704e924b5c8978923f) and they were replaced in the following "Latest Results" commit.
> 
> Do you have any thoughts? See attached screenshot of the most recent artifact from [259f732](https://github.com/NREL/resstock/pull/962/commits/259f732490bf123084f2be704e924b5c8978923f).
> 
> Screen Shot 2023-06-24 at 8 44 00 AM

@afontani The set of samples has changed. Looks like sampling is a [function of the options_lookup](https://github.com/NREL/resstock/blob/develop/resources/run_sampling_lib.rb#L23-L25).

.github/workflows/config.yml
```diff
@@ -251,7 +288,7 @@ jobs:
 
   update-results:
     runs-on: ubuntu-latest
-    needs: [integration-tests]
+    needs: [format-files,integration-tests]
```
joseph-robertson, line 291: 👍

.github/workflows/config.yml
```diff
@@ -300,6 +343,7 @@ jobs:
           cp -r results/upgrades/timeseries/*.csv test/base_results/upgrades/timeseries
           git add test/base_results
           git add docs
+          git add resources/options_lookup.tsv
```
joseph-robertson, line 346: Looks like the artifact download correctly places the options_lookup.tsv in the resources folder. So that's good - no copy necessary.

joseph-robertson (approved):
This looks good to me. Bummer we can't confirm simulation results don't change (because of the samples changing), but I don't see anything in this PR that would affect them anyway.

afontani:
> This looks good to me. Bummer we can't confirm simulation results don't change (because of the samples changing), but I don't see anything in this PR that would affect them anyway.

Yes. Agreed.

PR #975, opened 2022-08-11
Support auto_seasons arguments

joseph-robertson:
## Pull Request Description

Looks like OS-HPXML auto_seasons related arguments may have been updated, but ResStockArguments subsequently was not. I believe this was not caught because our testing project does not currently sample any setpoint options which enable the auto_seasons arguments.

## Checklist

Not all may apply:


shorowit:
Seems like a lot of code/complexity in ResStock just to support this. Would it make more sense to make the changes upstream? E.g., Update the heating/cooling season arguments in the BuildResHPXML measure to support a string like "BuildingAmerica".

joseph-robertson:
Probably.

afontani:
I can confirm this change works.

PR #979, opened 2022-08-17
Aspect ratio fix, try 2

joseph-robertson:
## Pull Request Description

I believe this had implications for ResStock post the "single-unit modeling" switch (and pre?). It was expected, with an aspect ratio of 1.8, that LR adiabatic surfaces had less area than FB outside surfaces. In reality, it seems the aspect ratio was 1 / 1.8 = 0.6 (the LR adiabatic surfaces had more area than FB outside surfaces). It's instead possible that we intended to model multifamily units with FB length < LR length, in which case the only bug here is having an incorrect definition for aspect ratio.

We were actually modeling something like this:
image
Is this what we want to continue modeling? If so, we need to change aspect ratio to 1 / 1.8 = 0.5556 in options lookup.

## Checklist

Not all may apply:


aspeake (approved):
Looks good. MF/SFA now follows the definition of aspect_ratio while maintaining the original modeling approach, w/ only negligible changes in results.

PR #982, opened 2022-08-24
Update join_tsv to use raw source_count

lixiliu:
## Pull Request Description

TSV changes from updating `join_distributions_in_order` to use raw source_weight (previously, it uses what is now called fallback_weight which artificially increases the weight of undersampled options, thus making it deviate more from the sample truth marginal distributions)

Companion resstock-estimation PR: https://github.com/NREL/resstock-estimation/pull/248

Comparison of Floor Area from AHS: [floor_area.xlsx](https://github.com/NREL/resstock/files/9428606/floor_area.xlsx)

## Checklist

Not all may apply:


.gitignore
```diff
@@ -7,6 +7,8 @@
 /testing_baseline
 /testing_upgrades
 /resources/buildstock.csv
+/project_*/**/*.csv
```
rajeee, line 10: Why are you ignoring .csv? I think this is causing the source_report.csv and options_saturations.csv to be deleted.

lixiliu: yeah, @afontani wanted only the tsvs to be in projects folder.

afontani (approved):
The marginal graphics is a great addition! Thanks.

PR #989, opened 2022-09-23
RTD reorganization

### LINKED ISSUE #882: Document outputs
The documentation doesn't seem to describe what outputs are produced by ResStock. For example, it may not be immediately clear to a user that Heating does not include Heating Fans/Pumps, or that the heating/cooling loads include duct losses. Maybe it's as easy as pointing to OpenStudio-HPXML output documentation plus describing any additional ResStock-specific outputs?

joseph-robertson:
## Pull Request Description

Closes #882.

TODO:
- [x] Update buildstockbatch wfg for `measures` coming _after_ HPXMLtoOpenStudio
- [x] Update run_analysis wfg for `measures` coming _after_ HPXMLtoOpenStudio
- [x] Finish `tutorial/architecture.rst`
- [x] Finish `outputs/index.rst`, `outputs/default_outputs.rst`, `outputs/optional_outputs.rst`
- [x] Remove `Advanced Tutorial / Installer Setup` section? Update: No, but updating language for its "optional" nature.

Proposal for the `Tutorial` section:
* `index.rst`
* `installation.rst`
* `architecture.rst`
  * renamed from `setup_analysis_project.rst`
  * remove measure arguments (this is already captured in bsb docs)
  * show general measure workflow order
  * point to measure xml descriptions
* `run_project.rst`
  * mostly left unchanged except for general comments about yml input file
* `conclusion.rst`

Proposal for the `Available Outputs` section:
* `index.rst`
* `default_outputs.rst`
  * new file
  * point to housing characteristics, simulation output, upgrade costs workflow outputs
  * **list outputs using on-the-fly generated bsb results.csv**
* `optional_outputs.rst`
  * new file
  * point to component loads, emissions, utility bills, QOIs, timeseries workflow outputs
  * **list outputs using on-the-fly generated bsb results.csv**

## Checklist

Not all may apply:


ekpresent:
This is feedback on documentation/tutorial/workflow_outputs.html.

1 - The introduction at the top of the page says 

"ResStock generates both a standard set of outputs as well as other outputs. The standard set of outputs include annual energy consumption, emissions, utility bills, etc., as well as optionally requested timeseries outputs. Other outputs include dwelling unit characteristics, upgrade cost information, and (optionally) quantities of interest." 

This is confusing and, I think, wrong. It makes it sound like there are "standard outputs" and 'other outputs" and I don't understand the difference. Why are building characteristics "other outputs" and not "standard outputs"?

I would prefer if we could split into 2-3 categories. 
A - outputs that will be there by default
B - (C) -  outputs that you can request, possibly split further by things that are inherently available (e.g. QOI, and I think unmet hours and component loads are in this category) versus things that you have to provide extra inputs to get (e.g. costs, emissions)

Basically I'd like to be able to understand "here's what you get", "here's what you can also get if you ask for it", and "here's what you can also get if you provide additional inputs"

2 - I'm a little unclear what the difference of intent is between this page and the one it links to, which seems to also be called workflow outputs. The one I am reviewing seems to have a dynamic list of the actual field names, while the linked one is more conceptual and actually has descriptions. Maybe this one should be called "available workflow outputs"?

3 - I think emissions should probably be done the way the "other upgrade cost information is" with wildcards? Or maybe not, maybe just say something like "Or other emissions outputs if choosing custom emissions inputs", because the list is the full range of what is available pre-computed - most people won't request all of them, and some people will request something else (by putting in their own input values).

ekpresent:
Comments on documentation/available outputs (and downstream)
- I like this format way better. Overall I have no major comments, the below are all minor.
- I would say "upgrade cost information" (not plural)
- I know we are trying to move away from "building", but I think an adjective would be helpful - "housing characteristics"? "dwelling unit characteristics"?
- What is "option level information"?
- a lot of these are grouped which isn't fully clear. Like you can either get all the elec end uses for an emissions scenario, or no elec information, but you can't get just the total. 

Comments on documentation/index
- Note at top of landing page: there is at least one dataset with an upgrade currently available! :)

ekpresent (commented):
I have reviewed the documentation portions of this PR and approve the merge.

PR #1041, opened 2023-03-03
Support skipping emissions calculations

joseph-robertson:
## Pull Request Description

[description here]

## Checklist

Not all may apply:


measures/ApplyUpgrade/measure.rb
```diff
@@ -382,15 +382,17 @@ def run(model, runner, user_arguments)
...
-    measures['BuildResidentialHPXML'][0]['emissions_scenario_names'] = values['emissions_scenario_names']
-    measures['BuildResidentialHPXML'][0]['emissions_types'] = values['emissions_types']
-    measures['BuildResidentialHPXML'][0]['emissions_electricity_units'] = values['emissions_electricity_units']
-    measures['BuildResidentialHPXML'][0]['emissions_electricity_values_or_filepaths'] = values['emissions_electricity_values_or_filepaths']
-    measures['BuildResidentialHPXML'][0]['emissions_fossil_fuel_units'] = values['emissions_fossil_fuel_units']
-    measures['BuildResidentialHPXML'][0]['emissions_natural_gas_values'] = values['emissions_natural_gas_values']
-    measures['BuildResidentialHPXML'][0]['emissions_propane_values'] = values['emissions_propane_values']
-    measures['BuildResidentialHPXML'][0]['emissions_fuel_oil_values'] = values['emissions_fuel_oil_values']
-    measures['BuildResidentialHPXML'][0]['emissions_wood_values'] = values['emissions_wood_values']
+    if values['emissions']
```
joseph-robertson, line 385: We need this switch, otherwise some args are set and others aren't.

rajeee:
Can we have emissions for other fuels even if electricity emission file is missing?

joseph-robertson:
@rajeee We cannot. An electricity factor is required. I talked briefly with @shorowit on this one. It's better to not calculate anything than to calculate partial emissions (emissions that only cover a part of the home's energy use), as that could be interpreted incorrectly.

As an alternative to skipping emissions calculations altogether, I suppose we could set an emission factor annual value for electricity (although I don't know what that value should/would be).

shorowit:
Note that nothing is stopping you from assigning an electricity emissions factor of zero, though I still think it's almost always a bad idea. But perhaps there could be a legitimate use case for it.

rajeee:
@shorowit  The reason we wanted to have this feature is because @afontani is adding Alaska and Hawaii to ResStock and we currently don't have any emissions file for AK and HI. So, although we can't get emissions for electricity in those regions, we could in theory get emissions for other fuels using the default emissions factor. I understand we could force behavior by setting emissions of zero for electricity.  But, yeah, maybe that's a bad approach to take.  Any thoughts on that @afontani ?

afontani:
I like not having values reported if we do not have a full set of inputs. We will need to figure out how to deal with AK and HI as there are no emissions data from Cambium for these states.

@rajeee : We might consider a BuildStock-Batch handling not existing values as NaNs instead of 0 for this field.

afontani (approved):

PR #1045, opened 2023-03-07
Support for shared water heating systems

### LINKED ISSUE #1044: Support for shared water heating systems
Currently we set `water_heater_num_units_served=1` without regard to what is sampled for "Water Heater In Unit" (Yes or No). Instead, should be using ResStockArguments to set `args['water_heater_num_units_served'] = Integer(args['geometry_building_num_units'])` when "Water Heater In Unit = No"?

joseph-robertson:
## Pull Request Description

Closes #1044.

## Checklist

Not all may apply:


afontani:
Looks like the argument `water_heater_num_units_served` distributes losses among the units served.
- https://github.com/NREL/OpenStudio-HPXML/blob/c84eecccef36347f6b85519d3b98682a597f13ea/BuildResidentialHPXML/measure.rb#L1820-L1825
- https://github.com/NREL/OpenStudio-HPXML/blob/7e9429ac8f288a22aabbd1d4b91aa0b781564522/HPXMLtoOpenStudio/resources/waterheater.rb#L1513-L1516

afontani:
Look into an argument related to distance to the furthest fixture.

joseph-robertson:
Meaning any time we have a shared water heater, we'd want to use a shared recirculation system?

shorowit:
@afontani See `PipingLength` here: https://openstudio-hpxml.readthedocs.io/en/latest/workflow_inputs.html#standard

joseph-robertson:
@afontani @shorowit What's the action item here? Do we want to increase (using ResStockArguments) piping length for units with shared water heaters? For example, recalculate the following but using _building_ `CFA` and _building_ `NCfl`?

```
PipeL = 2.0 * (CFA / NCfl)^0.5 + 10.0 * NCfl + 5.0 * Bsmnt
```

joseph-robertson:
> @afontani @shorowit What's the action item here? Do we want to increase (using ResStockArguments) piping length for units with shared water heaters? For example, recalculate the following but using _building_ `CFA` and _building_ `NCfl`?
> 
> ```
> PipeL = 2.0 * (CFA / NCfl)^0.5 + 10.0 * NCfl + 5.0 * Bsmnt
> ```

I tried this out, and put a plot in the description above. Now we are seeing more losses due to switching to shared water heaters vs a reduction in losses when maintaining the same pipe length.

afontani:
@shorowit mentioned describing the water heater and the size would be larger than a single unit. Autosizing routines are used for an in-unit water heater. Suggests not distributing tank losses. Maybe the piping length update could be kept.

@jmaguire1 : mentioned Ecotrope has a tool to size shared water heating systems. The recirculation loop typically runs continuously. Are we capturing the energy (look at the uncontrolled recirculation loop arguments)?

jmaguire1:
Trying to revive this one. I think what we're proposing right now isn't quite right though. When you have a shared water heater, the most common scenario is that there's a single plant (might be multiple WHs in series/parallel, but we'd probably model as always a single large tank) that serves the load in each unit. Since the WH might be far away from some units, it's common to have a recirculation loop from the central WH to each unit, which then connects to the in unit distribution systems to go from the recirc loop to each end use. It's common for these loops to run 24/7 so hot water is always available, but that also introduces a small continuous load on the water heater. Roughly looks like this:

image

A couple of issues we'd have to deal with:
- How do we introduce that recirc load onto the WH (apportioned to each unit)? We could assume a temperature loss and flow rate for the recirculation loop and use that to calculate the load from recirculation on the central WH and then apportion that to each unit. We'd want this to show up as an additional small but continuous hot water draw from the WH.
- What temperature does the tank see? If we're modeling multifamily there'll be a lot of units where the only zone is the living space, so there isn't a zone representing the mechanical room. 
- If it's a HPWH, I'm not sure this approach will work. Central HPWHs look pretty different than in unit ones, with different performance maps, pumped condensers, multiple tanks, etc. So I'd just exclude this for now, Joe and I are doing some work on explicit modeling of this for another project but it requires modeling the whole building at once.

PR #1056, opened 2023-03-15
Support sample_weight in buildstock.csv

joseph-robertson:
## Pull Request Description

Adds support for a `sample_weight` column in the precomputed buildstock.csv. By default, BuildExistingModel writes `build_existing_model.sample_weight` based on calculating `n_buildings_represented / n_datapoints` (this calculation is done in the workflow generator and then passed into the BuildExistingModel measure). Now, if `sample_weight` already exists in the buildstock.csv, it will write this value instead of the calculated one.

This PR also:
* removes registering of `weight` - it doesn't appear to be used anywhere
* removes the `number_of_buildings_represented` argument from BuildExistingModel - it appears it was only being used in the context of a PAT project, which we no longer support

## Checklist

Not all may apply:


joseph-robertson:
Should the new allowable column, for consistency, be "Sample Weight" instead of "sample_weight"?

measures/BuildExistingModel/measure.rb
```diff
@@ -600,8 +600,14 @@ def run(model, runner, user_arguments)
...
-    if args['sample_weight'].is_initialized
-      register_value(runner, 'weight', args['sample_weight'].get.to_s)
+    sample_weight = args['sample_weight'].is_initialized ? args['sample_weight'].get : nil
+    if bldg_data.keys.include?('sample_weight')
+      sample_weight = bldg_data['sample_weight']
+      register_value(runner, 'sample_weight', sample_weight.to_s)
+    end
+
+    if !sample_weight.nil?
+      register_value(runner, 'weight', sample_weight.to_s)
```
aspeake, line 610: Why are these registered as different variables? Should it just be one or the other gets stored? I do see there is a "weight" and a "sample_weight" in results.json, but then just "sample_weight" is output in results_up00.csv, can you help clarify this?

test/tests_yml_files/yml_precomputed_weight/testing_baseline.yml
```diff
@@ -0,0 +1,57 @@
...
+
+sampler:
+  type: precomputed
+  args:
+    sample_file: buildstock.csv
+
+workflow_generator:
+  type: residential_hpxml
+  args:
+    build_existing_model:
```
aspeake, line 16: What happens if we also add a `sample_weight` field here?

aspeake (commented):
Looks good, just had a couple of questions

measures/BuildExistingModel/measure.rb
nmerket: I think it gets registered so that it will show up in the outputs in results.csv. Then we can use it in downstream postprocessing.

joseph-robertson: But doesn't it [get removed](https://github.com/NREL/buildstockbatch/blob/develop/buildstockbatch/postprocessing.py#L168) in favor of using `sample_weight`?

nmerket: It would appear that it does.

joseph-robertson: Any recommendations what should be done here in regard to `weight`? Nothing? Or should I try removing it and see what happens?

test/tests_yml_files/yml_precomputed_weight/testing_baseline.yml
joseph-robertson: Do we want to support explicitly setting `sample_weight` from the yml file?

aspeake: No, I don't think that is necessary, bsb still passes `sample_weight` just not explicitly from the user, so I'm wondering what happens when sample_weight gets passed as an argument and there is a sample_weight column in the buildstock. Seems like the buildstock one overwrites it, right?

measures/BuildExistingModel/measure.rb
aspeake: I'm a little lost on where `sample_weight` gets used downstream, because it seems like `sample_weight` never got set here before, only `weight`

test/tests_yml_files/yml_precomputed_weight/testing_baseline.yml
joseph-robertson: That's correct. The workflow generator will continue to calculate `sample_weight`. But this PR will allow the presence of a `sample_weight` column in buildstock.csv to overwrite that calculated value.

measures/BuildExistingModel/measure.rb
joseph-robertson: `sample_weight` is a measure argument for BuildExistingModel. As long as this argument gets assigned a value, it will "register" as `build_existing_model.sample_weight` (i.e., you don't need the `register_value(runner, 'sample_weight', ...)` call). This PR "re-registers" (overwrites) the `sample_weight` value as it exists in the buildstock.csv.

PR #1058, opened 2023-03-16
Data dictionary + automated RTD generation

### LINKED ISSUE #1043: Data Dictionary of inputs and outputs
One of the pain points we experience when importing ResStock runs into SightGlass (resstock.nrel.gov) is that the outputs and format of the outputs from ResStock frequently change. This causes our data processing for that to break and require many hours of manual updating every time we go to bring new data in.

It would be really helpful to have a data dictionary of the outputs ResStock produces meaning every column name (including the input and output columns) in the results.csv and timeseries parquet files. It should also include some flags about which are end uses to include in the sum vs aggregates (net or totals), units, other random outputs like load, emissions, etc. To keep this in sync, it should be verified against the CI runs of ResStock and if there is a discrepancy you get a big ❌ on your checks.

@rajeee @afontani @trynthink @ekpresent @joseph-robertson

joseph-robertson:
## Pull Request Description

Closes #1043.

Introduce:
- [x] `resources/data/dictionary/inputs.csv`
  * This is a file created by hand
  * It contains every output that is prefixed with `build_existing_model`
  * Columns: Name, Description
- [x] `resources/data/dictionary/outputs.csv`
  * This is a file created by hand
  * Columns: Index, Row Index, Sums To, Annual Name, Annual Units, Timeseries ResStock Name, Timeseries BuildStockBatch Name, Timeseries Units, Notes
- [x] New tests in `test/test_bsb_analysis.rb`:
  * `test_<project>_inputs`: check characteristics columns against the dictionary
  * `test_<project>_annual_outputs`: check annual columns against the dictionary, and check Sums To column
  * `test_timeseries_resstock_outputs`: check ResStock timeseries columns against the dictionary, and check Sums To column
  * `test_timeseries_buildstockbatch_outputs`: check BuildStockBatch timeseries columns against the dictionary, and check Sums To column
- [x] `docs/read_the_docs/source/workflow_outputs/csv_tables.rb`
  * creates individual csv tables, organized by category, that RTD renders
- [x] `docs/read_the_docs/source/workflow_inputs/create_characteristics_rst.rb`
  * creates Housing Characteristics section of RTD; subsections are each Parameter referenced from Available Outputs 

Checklist:
- [x] Timeseries names reported by bsb vs those in `results_timeseries.csv`; could we include the bsb-processed timeseries names (e.g., "Timeseries BuildStockBatch Name") in addition to the raw datapoint timeseries names (e.g., "Timeseries ResStock Name")? Yes.
  * `parquet/timeseries/upgrade=0/groupX.parquet` (use pandas for reading these files)
- [x] Create a second `build_existing_model.` table with the argument rows (i.e., the blank Options rows).
- [x] Explicitly mention in the docs that you need to look at the raw data dictionary to understand how outputs relate to each other (i.e., "Row Index" and "Sums To" columns).
- [x] In Upgrade Costs, specify that number of options go from 1..25 (near "represents any of the defined option numbers.").
- [x] Update Housing Characteristics section based on the image below (this may be some kind of `source_report.csv` to `.rst` exercise); requires that `source_report.csv` is updated to contain columns of interest; create a placeholder "Description" column until it can be populated.
- [x] Are RTD docs actually being automatically generated? They are for the docs artifact, but what about committing updated `characteristics.rst` and `csv_tables/*.csv` back to the repository? Should probably move the "Build documentation" step back to the "update-results" job so we can commit changes to the `docs` folder.
- [x] Talk with Noel about whether we want, e.g., `<ScenarioName>` instead of actual emissions scenario. Update: yes, we should use `<scenario_name>` (and `<type>`) in the data dictionary file, for both emissions and utility bills.

image

## Checklist

Not all may apply:


resources/data/dictionary/outputs.csv
```diff
@@ -0,0 +1,377 @@
...
+63,report_simulation_output.emissions_co_2_e_lrmer_mid_case_15_electricity_heating_fans_pumps_lb,lb,Emissions: CO2e: LRMER_MidCase_15: Electricity: Heating Fans/Pumps,lb,84,Scenario emissions for this Electricity end use only (one row per end use)
+64,report_simulation_output.emissions_co_2_e_lrmer_mid_case_15_electricity_heating_heat_pump_backup_lb,lb,Emissions: CO2e: LRMER_MidCase_15: Electricity: Heating Heat Pump Backup,lb,84,Scenario emissions for this Electricity end use only (one row per end use)
+65,report_simulation_output.emissions_co_2_e_lrmer_mid_case_15_electricity_heating_lb,lb,Emissions: CO2e: LRMER_MidCase_15: Electricity: Heating,lb,84,Scenario emissions for this Electricity end use only (one row per end use)
+66,report_simulation_output.emissions_co_2_e_lrmer_mid_case_15_electricity_hot_tub_heater_lb,lb,Emissions: CO2e: LRMER_MidCase_15: Electricity: Hot Tub Heater,lb,84,Scenario emissions for this Electricity end use only (one row per end use)
```
shorowit, line 67: I don't know what the right answer is, but it seems misleading to have "lrmer_mid_case" as part of the name here since it will be specific to which scenario(s) are in the yaml.

Also, all the "(one row per end use)" should be removed since you are already enumerating all of the end uses here.

joseph-robertson: Yeah, I've thought about the scenario name issue, too. I put an attempted [clarifying note](https://github.com/NREL/resstock/blob/data-dictionary/docs/read_the_docs/source/outputs/optional_outputs.rst#emissions) in the docs about it.

Good catch on the "one row per end use" thing.

nmerket:
Just looking at [this file](https://github.com/NREL/resstock/blob/data-dictionary/resources/data/dictionary/outputs.csv), this is _awesome_. We'll have to change some things in our downstream processing, but it'll be so much easier now. Thank you.

joseph-robertson:
After meeting:
- [ ] Describe "Sums To" in documentation
- [ ] Change emissions/bills to use `<place_holders>`
- [ ] Provide more links between ResStock docs and bsb docs

joseph-robertson:
@rajeee Where, on [the bsb hpxml wfg docs page](https://buildstockbatch.readthedocs.io/en/stable/workflow_generators/residential_hpxml.html), do you suppose we should link to the resstock docs?

joseph-robertson:
> @rajeee Where, on [the bsb hpxml wfg docs page](https://buildstockbatch.readthedocs.io/en/stable/workflow_generators/residential_hpxml.html), do you suppose we should link to the resstock docs?

@rajeee and I talked a bit about this. The references should be within the applicable sections of the wfg docs page. We should get this PR merged before making the wfg docs updates, as I need the exact urls.

afontani (approved):
This is good work. Let's get it merged.

PR #1060, opened 2023-03-22
Latest OS-HPXML

joseph-robertson:
## Pull Request Description

Pulls in:
- [x] https://github.com/NREL/OpenStudio-HPXML/pull/1321
  - Set `air_leakage_multifamily_value_type=guarded test` in ResStockArguments

Checked means that something in resstock was changed or updated as a result of pulling in.

## Checklist

Not all may apply:


measures/ResStockArguments/measure.rb
```diff
@@ -640,6 +640,7 @@ def run(model, runner, user_arguments)
 
       # Apply adjustment to infiltration value
       args['air_leakage_value'] *= exposed_wall_area_ratio
+      args['air_leakage_multifamily_value_type'] = HPXML::InfiltrationTestGuarded
```
shorowit, line 643: Why not specify in options_lookup.tsv?

joseph-robertson: Certainly could. Seems more appropriate alongside all the infiltration adjustment stuff, no?

shorowit: No, the adjustments are not converting from compartmentalization values to guarded values. They are just increasing some units slightly and decreasing others, while keeping the overall infiltration for the MF building unchanged.

joseph-robertson: But we still want to set to "guarded test", right? Seems like putting that here implies were not unintentionally missing setting it to "compartmentalization test".

PR #1067, opened 2023-04-06
ServerDirectoryCleanup modeler_description

joseph-robertson:
## Pull Request Description

I'm pretty sure that by actually setting this string, we can avoid the measure.xml from being modified every time `update_measures` is called.

## Checklist

Not all may apply:


.github/workflows/config.yml
```diff
@@ -195,7 +195,7 @@ jobs:
       - name: Compare samples and results
         run: |          
           pip install numpy
-          pip install pandas
+          pip install pandas==1.5.3
```
joseph-robertson, line 198: Fixing this version to <2.0 due to seeing some weird/different stuff with `df.read_csv()`.

PR #1071, opened 2023-05-08
Add capacity retention arguments for ASHP

joseph-robertson:
## Pull Request Description

Similar to https://github.com/NREL/resstock/pull/898, but for ASHP.

I tested this by (temporarily) adding a new option in the lookup that sets values for `heat_pump_capacity_retention_temp` and `heat_pump_capacity_retention_fraction`. Then in `national_upgrades.yml` I created a new upgrade that sets the new option. When I run `run_analysis.rb` for a datapoint (with the same base ASHP), I can see that the IDF files differ by their (1) Gross Rated Heating COP and (2) "Heat-CAP-fT1" biquadratic curve:

image
image

## Checklist

Not all may apply:


shorowit (approved):
Implementation looks good. I also reproduced your test but set the sizing methodology for the ASHP upgrade to MaxLoad and correctly saw the autosized heat pump capacity increase in the HPXML/IDF. 👍

PR #1093, opened 2023-06-26
HVAC Secondary Heating, try 2

joseph-robertson:
## Pull Request Description

Companion PR: https://github.com/NREL/resstock-estimation/pull/375

Support options sampled from `HVAC Secondary Heating xxx`. Update the ResStockArguments measure to adjust the primary system's fraction of heat load served such that the sum of fractions does not exceed 1.0, for the following:
  * when `heating_system_type!=none`
  * when `heat_pump_type!=none`

This PR also pulls in the develop branch of OS-HPXML containing related fixes in https://github.com/NREL/OpenStudio-HPXML/pull/1414.

## Checklist

Not all may apply:

  - [ ] If related to resstock-estimation, checklist includes data dictionary, source report, options_lookup


measures/ResStockArguments/measure.rb
```diff
@@ -496,6 +496,25 @@ def run(model, runner, user_arguments)
       args[:air_leakage_has_flue_or_chimney_in_conditioned_space] = true
     end
 
+    # HVAC Secondary
+    if args[:heating_system_2_type] != 'none'
+      if args[:heating_system_type] != 'none'
+        if ((args[:heating_system_fraction_heat_load_served] + args[:heating_system_2_fraction_heat_load_served]) > 1.0)
+          info_msg = "Adjusted fraction of heat load served by the primary heating system (#{args[:heating_system_fraction_heat_load_served]}"
+          args[:heating_system_fraction_heat_load_served] -= (args[:heating_system_2_fraction_heat_load_served] - (1.0 - args[:heating_system_fraction_heat_load_served]))
```
rajeee, line 504: Assuming f1=heating_system_fraction_heat_load_served, and f2 = heating_system_2_fraction_heat_load_served, the logic you have here is essentially:

```
f1 -= (f2 - (1 - f1))
f1 = f1 - (f2 - (1 - f1))
f1 = f1 - (f2 -1 + f1) 
f1 = f1 - f2 + 1 - f1 
f1 = 1 - f2
```
I think it's easier to follow if you instead change the code to:

```suggestion
          args[:heating_system_fraction_heat_load_served] = 1 - args[:heating_system_2_fraction_heat_load_served]
```
And do the same for the heat_pump case.

joseph-robertson: Hmm, I think they're both wrong. Say f1 was 0.6 and f2 was 0.25. Then in either case you're changing f1 to 0.75, which we don't want.

joseph-robertson: Oh wait. The `if` that this is inside of protects us against that.

PR #1113, opened 2023-08-01
Add resilience arguments to yml files

joseph-robertson:
## Pull Request Description

Include the new `include_timeseries_resilience` argument in example yml files.

## Checklist

Not all may apply:

  - [ ] ~If related to resstock-estimation, checklist includes data dictionary, source report, options_lookup~


jmaguire1:
One thought/suggestion for resilience: there's a pretty non obvious thing you should change when looking at resiliency. There's a "temperature capacitance multiplier" (TCM) input in E+ that increases the thermal mass of the air node. If you just stick with the default we have right now (1.0), you'll see the building temperature increase/decrease faster than appears realistic. Looking at the literature, when people are calibrating models there isn't much consensus on what value should be used, there's a range between 3-15 [depending on the building](https://docs.google.com/spreadsheets/d/1ebSmvDFdXEXVRdvkzqMF1C9MwHrHCQKFF75QMkPgd7A/edit#gid=0). I've been using a value of 7.0, based on what's worked well in the projects we've run, where we ran HIL experiments in SPL and tried to calibrate to a real building by changing the TCM.

We have an input for the temperature capacitance multiplier right now, and the value gets set here: https://github.com/NREL/resstock/blob/98d5baba9efc9f1a9a6c9bed2d90fdf30de6dfc4/resources/hpxml-measures/HPXMLtoOpenStudio/resources/simcontrols.rb#L39. Would we ever want to default to a value of 7.0 (or something > 1) whenever resilience outputs are requested? I feel like most users wouldn't know to adjust the multiplier.

I believe we didn't change our default TCM because it led to some failures in some tests? @shorowit do you remember exactly what tests were failing? I vaguely recall that it was only a 1-2% difference in annual energy consumption but that was just barely enough to flip a few tests from passing to failing.

shorowit:
@jmaguire1 Why would the battery resilience metric be significantly affected by the temperature capacitance multiplier? We're not modeling power outages to calculate it, we're just seeing how long the battery would last under normal electricity usage conditions.

PR #1146, opened 2023-10-12
RTD: summarize lookup arguments

joseph-robertson:
## Pull Request Description

This PR:
* Adds remaining arguments (i.e., `<arg_name>=auto`) to `options_lookup.tsv` so that **every ResStockArguments argument is included**.
* Adds a unit test to ensure that _all_ ResStock arguments are called out in `options_lookup.tsv`.
* By housing characteristic parameter, automates summarizing **Arguments** sets used in `options_lookup.tsv` (including Name, Required, Units, Type, Choices, Description). See below for example.
* Adds a `README.md.erb` to all ResStock measures so that `README.md` documentation is automatically generated.
* Adds introduction in `Housing Characteristics` for describing subsections and the "auto" argument choice. See below.

---

image

---

image

## Checklist

Not all may apply:

  - [ ] ~If related to resstock-estimation, checklist includes [data dictionary](https://github.com/NREL/resstock/tree/develop/resources/data/dictionary), [source report](https://github.com/NREL/resstock/tree/develop/project_national/resources/source_report.csv), [options saturation](https://github.com/NREL/resstock/tree/develop/project_national/resources/options_saturations.csv), [options_lookup](https://github.com/NREL/resstock/blob/develop/resources/options_lookup.tsv).~


shorowit:
I realize you've already gone done a certain path and I'm coming to it late, but I can't help but wonder if it would be simpler to just include every argument in the options_lookup.tsv somewhere. Could even write a test to make sure they are all covered.

shorowit:
Related: https://github.com/NREL/OpenStudio-HPXML/issues/1518

joseph-robertson:
Suggestion from @afontani: We probably want to link to argument descriptions.

docs/read_the_docs/source/workflow_inputs/create_characteristics_rst.rb
```diff
@@ -46,4 +79,44 @@ def to_underscore_case
...
+  f.puts('*' * name.size)
+  f.puts
+  f.puts('.. list-table::')
+  f.puts('   :header-rows: 1')
+  f.puts
+  f.puts('   * - Name')
+  f.puts('     - Units')
+  f.puts('     - Choices')
+  f.puts('     - Description')
+  r_arguments.sort.each do |r_argument|
```
shorowit, line 107: Why sort? Seems like it would be better to preserve the order in the measure. For example, it'd be nice if `clothes_washer_present` was the first clothes washer argument listed.

joseph-robertson: The order in `r_arguments` is determined by the order (from left to right) in `options_lookup.tsv`. But I could probably sort them according to the order in the measure.xml...

joseph-robertson:
We need to better orient readers of the Housing Characteristics page to the idea of "auto" and what that means, how things connect to OS-HPXML documentation and HPXML fields, etc.

We should have a paragraph up front on "auto", and then for each ResStock parameter we should have link(s) to the related OS-HPXML documentation sections (e.g., `HVAC Heating Efficiency` -> [HPXML Heating Systems](https://openstudio-hpxml.readthedocs.io/en/latest/workflow_inputs.html#hpxml-heating-systems), [HPXML Heat Pumps](https://openstudio-hpxml.readthedocs.io/en/latest/workflow_inputs.html#hpxml-heat-pumps)).

docs/read_the_docs/source/workflow_inputs/characteristics.rst
```diff
@@ -3,6 +3,27 @@
...
+- **Created by**
+- **Source**
+- **Assumption**
+
+Additionally for each parameter, an **Arguments** table is populated (if applicable) based on the contents of `ResStockArguments's measure.xml file <https://github.com/NREL/resstock/blob/develop/measures/ResStockArguments/measure.xml>`_:
+
+- **Name**
+- **Units**
+- **Choices**
+- **Description**
```
joseph-robertson, line 19: Look at BuildResHPXML/measure.xml to get: either (A) **Choices** gets updated with "auto" and "<type>" or (B) **Type**, and then (C) **Required** (Yes or No, so that we can populate "auto" when it's not a choice argument). Optional here is to populate "true" and "false" in the Choices column for bool arguments.

@afontani

docs/read_the_docs/source/workflow_inputs/characteristics.rst
```diff
@@ -3,6 +3,27 @@
 Housing Characteristics
 =======================
 
+Each parameter sampled by the national project is listed alphabetically below.
+For each, the following (if applicable) are reported based on the contents of `source_report.csv <https://github.com/NREL/resstock/blob/develop/project_national/resources/source_report.csv>`_:
+
+- **Description**
+- **Created by**
+- **Source**
+- **Assumption**
```
joseph-robertson, line 12: These should be semi-colon delimited. So we should separate these out into bullets.

docs/read_the_docs/source/workflow_inputs/create_characteristics_rst.rb
```diff
@@ -46,4 +112,46 @@ def to_underscore_case
   write_subsection(f, row, 'Created by', '*')
   write_subsection(f, row, 'Source', '*')
   write_subsection(f, row, 'Assumption', '*')
```
joseph-robertson, line 114: Assumption**s**? Handle this on the resstock-estimation side. These should probably not be hard-coded here.

docs/read_the_docs/source/workflow_inputs/characteristics.rst
joseph-robertson: I'm writing the following columns to the Arguments table, in the following order:
* Name
* Required (comes from BuildResidentialHPXML)
* Units
* Type (comes from BuildResidentialHPXML)
* Choices
  * all optional choice arguments have "auto" in addition to all other choices
  * all optional boolean arguments have "auto", "true", "false"
  * add "auto" choice if "OS-HPXML default" string in the Description and Type is String/Double/Integer
* Description

joseph-robertson:
Need to either (A) wait for source_report formatting updates, or (B) fix some of the list item indentation issues.

PR #1193, opened 2024-01-19
add Census dwelling unit count table as a resource

lixiliu:
## Pull Request Description

I keep finding myself having to look up the latest dwelling unit count information when creating new runs with different geography needs. We should just keep a table of dwelling unit count by state from various vintage of ACS 5 years (2018-2022).

Data source: https://data.census.gov/table

## Checklist

Not all may apply:

  - [ ] If related to resstock-estimation, checklist includes [data dictionary](https://github.com/NREL/resstock/tree/develop/resources/data/dictionary), [source report](https://github.com/NREL/resstock/tree/develop/project_national/resources/source_report.csv), [options saturation](https://github.com/NREL/resstock/tree/develop/project_national/resources/options_saturations.csv), [options_lookup](https://github.com/NREL/resstock/blob/develop/resources/options_lookup.tsv).


afontani:
@lixiliu : Thinking about this a little. I have a few thoughts:
- This information is available with our spatial lookup tables and then scaled to the `n_buildings_represented` value in the project yaml or the ACS vintages.
- If the above resource is not good enough for your uses, I would like for this file to get automatically generated in CI and committed in the "latest commit".
- Also why State level, what about for every "County and PUMA", this way County, PUMA, and State Counts can easily be aggregated.

Let me know what path you prefer.

lixiliu:
> @lixiliu : Thinking about this a little. I have a few thoughts:
> 
> * This information is available with our spatial lookup tables and then scaled to the `n_buildings_represented` value in the project yaml or the ACS vintages.
> * If the above resource is not good enough for your uses, I would like for this file to get automatically generated in CI and committed in the "latest commit".
> * Also why State level, what about for every "County and PUMA", this way County, PUMA, and State Counts can easily be aggregated.
> 
> Let me know what path you prefer.

Summary of our discussion offline: The table is meant as a lookup for `n_buildings_represented`, not duplicating our spatial tables or tsvs. A "resources" section will be added to README to highlight this file as well as other useful things we've been adding to ResStock to enhance understanding or workflow.

afontani (changes requested):
Can we move the documentation to the Read-the-Docs instead of the Readme?

docs/read_the_docs/source/workflow_inputs/characteristics.rst
```diff
@@ -1,6 +1,6 @@
 .. _housing_characteristics:
 
-Housing Characteristics
+Housing Characteristics Distributions
```
joseph-robertson, line 3: I think you need to make this change instead in `docs/read_the_docs/source/workflow_inputs/create_characteristics_rst.rb` -- this script automatically generates characteristics.rst.

joseph-robertson (commented):
What's the plan/status for this PR?

Should we open a new ticket for addressing the Emissions and Utility Bills updates in the future? (Note that the Workflow Outputs section already contains Emissions and Utility Bills subsections.)

Issue #1161, opened 2023-11-02
Temperature Capacitance Multiplier default values


jmaguire1:
A little background for everyone:

EnergyPlus has a "temperature capacitance multipler" (TCM) input that affects the thermal mass of the building. While we explicitly model massive objects like furniture and interior walls, the temperature capacitance multiplier applies directly to the air node in the space and increases it's capacitance by the TCM. We've noticed if you simulate quick temperature changes, either due to setup/setback or starting an outage, the indoor temperature drops quicker than we'd expect with a TCM of 1.0, the default.

If you look at the literature, others have noticed this and in model validation come up with values ranging from [3.6-15](https://docs.google.com/spreadsheets/d/1ebSmvDFdXEXVRdvkzqMF1C9MwHrHCQKFF75QMkPgd7A/edit#gid=0). It's a pretty wide range which is implying that the mass is somewhat building specific. We also think that part of why you might need a TCM > 1 is that we're not explicitly accounting for the duct volume, which does have some impact on how quickly the thermostat sees heating/cooling reach the thermostat. It's hard with the data available to disentangle this impact from the potential of missing some thermal mass in the buildings.

We've been defaulting to a value of 7.0 in [OCHRE ](https://github.com/NREL/OCHRE/blob/79fd6ad9d3251347886bb91e88f7ce44fc00e376/ochre/Models/Envelope.py#L197C20-L197C20). 7.0 is based on HIL experiments Bethany and Sugi performed with a lab home (simulated occupants, so basically everything was known) when they tried to calibrate the home model to the field data. It's also close to the middle of the range seen in the literature.

In cases with no outage/constant setpoints, I'd expect next to no impact (like 1-2 % difference in annual whole home energy). The impact will be a little bigger if there's setup/setback, but also still pretty small annually. We might underestimate how much load can be shifted if we're underestimating the thermal mass though.  This matters quite a bit more if you have either an outage or the house temperature is floating (ie if it has no cooling in the summer). It seems like we have more global cooling measures and some projects are starting to look at modeling outages, so I want to raise the issue that we might want to change the default value so people don't have to worry about this in those use cases.

A couple questions I think we should discuss:
[...]

PR #1223, opened 2024-03-26
Update Temperature Capacitance Multiplier to 7.0

afontani:
## Pull Request Description

Closes: #1161 

After discussion and recent investigations into better understanding the resilience impacts of the ResStock timeseries results for power outages and load-shifting events, there is some evidence that the amount of thermal mass in ResStock is under-predicted. A quick (although non-physical fix) is to increase the temperature capacitance multiplier. A recommended number of 7.0 has been suggested by @jmaguire1. There are multiple efforts currently trying to leverage ResStock in these applications. While work is being done to better figure out a better solution, a quick fix is implemented to get the results in the acceptable range.

See related discussion in the attached issue for more details. 

## Checklist

Not all may apply:

  - [ ] If related to resstock-estimation, checklist includes [data dictionary](https://github.com/NREL/resstock/tree/develop/resources/data/dictionary), [source report](https://github.com/NREL/resstock/tree/develop/project_national/resources/source_report.csv), [options saturation](https://github.com/NREL/resstock/tree/develop/project_national/resources/options_saturations.csv), [options_lookup](https://github.com/NREL/resstock/blob/develop/resources/options_lookup.tsv).


resources/hpxml-measures/HPXMLtoOpenStudio/resources/hpxml_defaults.rb
```diff
@@ -142,7 +142,9 @@ def self.apply_header(hpxml_header, epw_file)
     end
 
     if hpxml_header.temperature_capacitance_multiplier.nil?
-      hpxml_header.temperature_capacitance_multiplier = 1.0
+      # Different than HPXML defaults, based on work from Jeff Maguire and Yueyue
+      # Discussion located in issue 1161 and will likely need to get updated later.
+      hpxml_header.temperature_capacitance_multiplier = 7.0
```
joseph-robertson, line 147: FYI there is a BuildResHPXML measure argument for this: https://github.com/NREL/OpenStudio-HPXML/blob/master/BuildResidentialHPXML/measure.rb#L114-L117.

We could either control it using the yml file (requires bsb updates) or using the options lookup (like `simulation_control_daylight_saving_enabled`).

afontani: @joseph-robertson : Thanks, I like the options_lookup suggestion. Are you able to revert this and then stub it out in options_lookup. This way it will be easy to change the value and when we all decide as a group what value to use, then we can switch it back to auto.

jmaguire1 (commented):
That's pretty much how I would have done it, looks good! Still investigating the exact number we should use but 7 is better than 1 for sure.

jmaguire1 (approved):

afontani:
A couple screenshots from the datapoint comparisons...
Screenshot 2024-03-29 at 10 59 35 AM
Screenshot 2024-03-29 at 11 00 32 AM
Screenshot 2024-03-29 at 11 02 27 AM
Screenshot 2024-03-29 at 11 02 52 AM
Screenshot 2024-03-29 at 11 03 32 AM

resources/hpxml-measures/HPXMLtoOpenStudio/resources/hpxml_defaults.rb
```diff
@@ -142,7 +142,7 @@ def self.apply_header(hpxml_header, epw_file)
     end
 
     if hpxml_header.temperature_capacitance_multiplier.nil?
-      hpxml_header.temperature_capacitance_multiplier = 1.0
+      hpxml_header.temperature_capacitance_multiplier = 7.0
```
joseph-robertson, line 145: Recall I made a commit where we set this to 7.0 using the ResStockArguments measure; I don't think we need this change to `hpxml_defaults.rb`.

joseph-robertson:
Superseded by https://github.com/NREL/resstock/pull/1247.

Issue #1236, opened 2024-05-01
Hot water usage seems low


jmaguire1:
We recently made a change to hot water usage multiplier to try to get a more realistic spread in hot water usage (#1210), which gives much more reasonable variability but we're realizing had an unintended effect on the mean. The original distribution was 25% of homes had multipliers of 0.5, 25% of homes had a multiplier of 2.0, and the remaining 50% had a multiplier of 1.0, which led to an overall multiplier of 1.15. With the recent change, we have an average multiplier of 0.86 (see distribution below). The combined impact of this change is our average hot water usage decreased by 1.15/0.86 = 33% In a 30k sample run I did recently, we're seeing an average hot water usage of ~35 gal/day (after filtering out vacant units). I'm digging into good data sources for what national average number we should try to target, but my first inclination is we should be pretty close to the draw volume used for medium use water heaters in the UEF test (so 55 gal/day):

![image](https://github.com/NREL/resstock/assets/8376057/f324273d-d59f-4945-9d65-1ede1a66c6bb)

Thanks to Lixi and Rajendra for helping me figure this out and trying to work out the appropriate change here.

![image](https://github.com/NREL/resstock/assets/8376057/5bc1d8e7-be48-4f9e-8302-047c292afd08)

afontani:
The related PR for the data-driven multipliers: https://github.com/NREL/resstock/pull/1210

shorowit:
We should coordinate/discuss this. We are looking to change the default OS-HPXML behavior for how hot water usage, as a function of number of occupants, is calculated, in which case you would ideally use multipliers with a weighted average of 1.

jmaguire1:
So I went back to the data from the field study and shifted it so we can keep a similar spread, but have a mean be 1.0 (basically option 2 Tony described):



![Image](https://github.com/NREL/resstock/assets/8376057/bd547dec-10df-42bf-9bda-5bfeefd88694)

This is what the distribution of hot water usage multipliers would look like. Unless anyone objects to this approach, I think I'll update this in this manner. This also means that we'll be getting mean values consistent with what Scott is planning to change in terms of changing the default OS-HPXML behavior (to match this paper: https://www.fsec.ucf.edu/en/publications/pdf/fsec-pf-464-15.pdf)

shorowit:
Related PR: https://github.com/NREL/OpenStudio-HPXML/pull/1761

PR #1265, opened 2024-07-08
Update Hot Water Fixtures tsv

lixiliu:
## Pull Request Description

Resolves: https://github.com/NREL/resstock/issues/1236
Companion ResStock-Estimation PR: https://github.com/NREL/resstock-estimation/pull/420

Mean-shifted distribution from 0.8 to 1.0
Specifically, the previous distribution is specified as a lognormal distribution with a transformed mean of 0.8 and a stdev 0.2.
New distribution is tabulated directly from mean-shifted input data (with mean=1.0 and a stdev ~ 0.236).
image

## Checklist

Not all may apply:

  - [x] If related to resstock-estimation, checklist includes [data dictionary](https://github.com/NREL/resstock/tree/develop/resources/data/dictionary), [source report](https://github.com/NREL/resstock/tree/develop/project_national/resources/source_report.csv), [options saturation](https://github.com/NREL/resstock/tree/develop/project_national/resources/options_saturations.csv), [options_lookup](https://github.com/NREL/resstock/blob/develop/resources/options_lookup.tsv).


jmaguire1 (commented):
Looks good to me. I definitely still want to test this with a ResStock run, but I also need to pull in a branch of the latest OS-HPXML to do this. I'll talk to Joe about if we can do the run with these as open PRs.

jmaguire1:
Results with these updates:

image

This is MUCH better than we used to be. If we wanted to get to exactly 1.0, I think we could shift the mean in this distribution to be 3% lower (the average across all usage levels in this 30k sample, which should be big enough to be representative). That said, I'd be fine with this as is, compared to the huge boost in hot water use we were doing. I suspect this will have a pretty minimal effect on energy use, maybe on the edges of some HPWH cases you'll end up with more element usage, but overall being slightly higher than 100% on average seems like a pretty minimal difference.

shorowit:
Hmm, what is the possible explanation for why things are off by 2-5%? If the distribution has a mean of 1.0 and we have enough samples, it seems like we should be closer.

Would it make sense to do a quick run where we only use the 100% Usage option and see if that matches better? That could help isolate whether the discrepancy is caused by the energy model or the distribution.

jmaguire1:
@shorowit: The mean of the % distributions is also 1.02, so I'm pretty confident if we use all 100% multipliers we'll get 1.00 exactly. It seems like this is either not enough samples (unlikely) or the distribution is just slightly off when we generated them. I can look at tweaking the bin weights to get this to average exactly 1.00 if we think it's worth the effort. I can tweak the distributions and we can do another 30k run if we think that's worth doing.

afontani:
@jmaguire1 : to double-check the shifting of the distribution to make sure it was done correctly and correspond to the ~2% discrepancy.

jmaguire1:
So now that this is fixed in resstock-estimation (see https://github.com/NREL/resstock-estimation/pull/420) and tested with a 30k sample, I think this is good?

Here's what I get on my sample for an average multiplier so this is documented in the PR:
image

PR #1269, opened 2024-07-11
ResStockArguments: convert args to double, integer

joseph-robertson:
## Pull Request Description

Avoids a bug that would be introduced by making `air_leakage_value` optional in https://github.com/NREL/OpenStudio-HPXML/pull/1760.
Per suggestion by @shorowit, create a method for automatically converting ResStockArguments argument data types based on original argument type.

## Checklist

Not all may apply:

  - [ ] ~If related to resstock-estimation, checklist includes [data dictionary](https://github.com/NREL/resstock/tree/develop/resources/data/dictionary), [source report](https://github.com/NREL/resstock/tree/develop/project_national/resources/source_report.csv), [options saturation](https://github.com/NREL/resstock/tree/develop/project_national/resources/options_saturations.csv), [options_lookup](https://github.com/NREL/resstock/blob/develop/resources/options_lookup.tsv).~


measures/ResStockArguments/measure.rb
```diff
@@ -783,6 +784,27 @@ def modify_setpoint_schedule(schedule, offset_magnitude, offset_schedule)
...
     return schedule
   end
+
+  def convert_args(model, args)
+    measures_dir = File.absolute_path(File.join(File.dirname(__FILE__), '../../resources/hpxml-measures'))
+    full_measure_path = File.join(measures_dir, 'BuildResidentialHPXML', 'measure.rb')
+    measure = get_measure_instance(full_measure_path)
+
+    args.each do |name, value|
+      measure.arguments(model).each do |arg|
```
shorowit, line 794: Yes, this was the general approach I was suggesting. Nice. This should definitely reduce bugs going forward.

But why are you going through all those complication to get the measure arguments? You can just pass them in. See next comment...

measures/ResStockArguments/measure.rb
```diff
@@ -389,6 +389,7 @@ def run(model, runner, user_arguments)
 
     # assign the user inputs to variables
     args = runner.getArgumentValues(arguments(model), user_arguments)
+    args = convert_args(model, args)
```
shorowit, line 392: Can't you just pass in the measure arguments like so?

```suggestion
    args = convert_args(arguments(model), args)
```

joseph-robertson: We want only the original arguments from BuildResidentialHPXML, no?

shorowit:
Also, would it make sense to go one step further? Just read in all arguments as string, which will significantly simplify the code [here](https://github.com/NREL/resstock/blob/927678b666fc4225761875bf4934f32825f7829d/measures/ResStockArguments/measure.rb#L44-L64) and then also handle choice arguments in your new method.

measures/ResStockArguments/measure.rb
shorowit: I guess I assumed there's no harm on having all the arguments, but maybe there's a nuance I'm missing. Feel free to disregard my comment if you don't think it's better/correct.

joseph-robertson: Maybe you're right. I'll try it.

joseph-robertson:
> Also, would it make sense to go one step further? Just read in all arguments as string, which will significantly simplify the code [here](https://github.com/NREL/resstock/blob/927678b666fc4225761875bf4934f32825f7829d/measures/ResStockArguments/measure.rb#L44-L64) and then also handle choice arguments in your new method.

I suspect this would conflict with how we automatically build our RTD based on measure.xml. Thoughts here @afontani?

measures/ResStockArguments/measure.rb
joseph-robertson: Tried it. One nuance is that `geometry_unit_cfa` gets turned into a string argument. But we can just handle that one separately.

joseph-robertson: Er wait. This doesn't work at all -- `measure(arguments)` changes all BuildResidentialHPXML double and integers arguments to string...

shorowit: Oh, right. Okay then I guess you have to do this.

joseph-robertson:
Oh boy; I think this change explodes runtime for ResStockArguments. See the measure-tests CI job...

measures/ResStockArguments/tests/resstock_arguments_test.rb
```diff
@@ -48,7 +48,7 @@ def test_create_geometry_envelope
     arg_name_prefixes = ['geometry', 'door', 'window', 'skylight']
     buildstock_csv_path = File.join(File.dirname(__FILE__), '../../../test/base_results/baseline/annual/buildstock.csv')
     building_ids = (1..(CSV.read(buildstock_csv_path).size - 1)).to_a
-    building_ids = building_ids.sample(0.8 * building_ids.size, random: Random.new(12345)) # 80% for runtime reduction
+    building_ids = building_ids.sample(0.75 * building_ids.size, random: Random.new(12345)) # 75% for runtime reduction
```
joseph-robertson, line 51: Trying to keep total CI time ~2 hrs.

shorowit:
> Oh boy; I think this change explodes runtime for ResStockArguments. See the measure-tests CI job...

Seems like you are calling `get_measure_instance` multiple times on the same measures from different methods. Why not just call it once for each measure and store it as a class instance variable?

shorowit:
One additional thought. There are places in the measure where we are getting a string from an optional value (e.g., [using `.to_s` here](https://github.com/NREL/resstock/blob/34fb6733dc429e8d8418b88f0425179df2bda0e8/measures/ResStockArguments/measure.rb#L455)). But doesn't ResStock now require all arguments to be provided? Seems like you could remove the need to use `.to_s` by changing [all of these arguments](https://github.com/NREL/resstock/blob/34fb6733dc429e8d8418b88f0425179df2bda0e8/measures/ResStockArguments/measure.rb#L50-L55) to be required instead of optional?

shorowit:
And why are some of the ResStockArguments arguments optional? Like [this one](https://github.com/NREL/resstock/blob/34fb6733dc429e8d8418b88f0425179df2bda0e8/measures/ResStockArguments/measure.rb#L110)?

joseph-robertson:
> > Oh boy; I think this change explodes runtime for ResStockArguments. See the measure-tests CI job...
> 
> Seems like you are calling `get_measure_instance` multiple times on the same measures from different methods. Why not just call it once for each measure and store it as a class instance variable?

You mean introduce a new `initialize` method to the `ResStockArguments` class?

joseph-robertson:
> And why are some of the ResStockArguments arguments optional? Like [this one](https://github.com/NREL/resstock/blob/34fb6733dc429e8d8418b88f0425179df2bda0e8/measures/ResStockArguments/measure.rb#L110)?
> 
> EDIT: Maybe they should all be required and use "auto" if they are to be defaulted, so that _all_ arguments are consistent.

That one is optional because it's not required for SFD. When MF, it's sampled as either Bottom, Middle, or Top.

measures/ResStockArguments/tests/resstock_arguments_test.rb
shorowit: You should be able to revert this, right? I assume there's no longer an impact on runtime since you are reusing the measure instance.

shorowit: In fact, it should be faster now, if anything, since you are avoiding some duplicate calls to get measure instances.

joseph-robertson: Appears CI time still long for measure-tests. Apparently something newly introduced, probably in `convert_arguments`, over-shadows any savings from measure instances. Any ideas here?

docs/read_the_docs/source/changelog/changelog_dev.rst
```diff
@@ -6,6 +6,22 @@ Development Changelog
     :version: v3.3.0
     :released: pending
 
+    .. change::
+        :tags: workflow, bugfix
```
shorowit, line 10: Doesn't really seem like a bugfix to me. An existing ResStock user couldn't hit an issue, right?

docs/read_the_docs/source/workflow_inputs/characteristics.rst
```diff
@@ -62330,7 +62330,7 @@ Arguments
      - false
      - 
      - Integer
-     -
+     - auto
```
shorowit, line 62333: Does it make sense that these two `auto`'s were added here?

shorowit (approved):
Overall this looks good to me (assuming the CI is happy with my last commit). Though I do have a couple minor comments.

docs/read_the_docs/source/changelog/changelog_dev.rst
joseph-robertson: It fixes the impending bug as described above in the PR description, no?

docs/read_the_docs/source/workflow_inputs/characteristics.rst
joseph-robertson: Added `auto` for:
* `year_built`: `auto` is a valid option in the lookup; in fact, we use it in every Vintage option
* `geometry_unit_num_occupants`: `auto` is supported [here](https://github.com/NREL/resstock/blob/develop/measures/ResStockArguments/measure.rb#L464) per ResStockArguments

docs/read_the_docs/source/changelog/changelog_dev.rst
shorowit: Prior to pulling in the latest OS-HPXML, there was no bug. After pulling in the latest OS-HPXML and including this PR, there's no bug. Just seems like a back-end refactoring that won't affect any ResStock users.

joseph-robertson: I understand what you're saying; it's not worth mentioning as a bug if it's technically not merged into the code yet. Can you suggest a tag?

shorowit: Yep, that is my point. I'd call it a refactor.

PR #1312, opened 2024-10-17
Assign auto instead of 0 to mean unused

### LINKED ISSUE #1289: ResStock Args consistency with OS-HPXML
Compiling a list of inconsistencies between the ResStock Arguments in Options_lookup and OS-HPXML documentation:
- For `Water Heater Efficiency`, water_heater_recovery_efficiency=0 for all electric tank options such as "[Electric Premium](https://github.com/NREL/resstock/blob/91f6f56a823e035e583dd5b78468ea19b248f79d/resources/options_lookup.tsv#L13899-L13900)" and heat pump. In [OS-HPXML](https://openstudio-hpxml.readthedocs.io/en/latest/workflow_inputs.html#id1195), all electric storage WH has RE=0.98.

### LINKED ISSUE #1293: Confusing use of `foo=0` for unused variables in options_lookup.tsv
ResStock uses `foo=0` when a variable does not apply in options_lookup.tsv. That is confusing because there's no way to really differentiate between a variable that is using zero vs a variable that is not being used.

Can we change those that are not used to something like `foo=na`? (Or simply `foo=`?)

joseph-robertson:
## Pull Request Description
It's confusing to assign 0 to an argument to mean unused. So we'll:
* still assign 0 if it makes sense as an input (even if unused)
* assign none when 0 doesn't make sense as an input

### Related Pull Requests

### Related Issues
Closes #1289.
Closes #1293.

#### Required:
- [ ] [Technical reference documentation](https://github.com/NREL/resstock/tree/develop/docs/read_the_docs) has been updated


shorowit:
"We'll leave 0 when: the argument is required"

That's a big gap that will still result in situations where the input is confusing or misleading. In these cases, couldn't we automatically remap "auto" to a valid value [here](https://github.com/NREL/resstock/blob/0081eb72e87480d8c11fdf5fe99e07fc757a58e2/measures/ResStockArguments/measure.rb#L838-L839)? Or maybe we should reconsider whether some of the BuildResHPXML arguments should be required.

joseph-robertson:
> "We'll leave 0 when: the argument is required"
> 
> That's a big gap that will still result in situations where the input is confusing or misleading. In these cases, couldn't we automatically remap "auto" to a valid value [here](https://github.com/NREL/resstock/blob/0081eb72e87480d8c11fdf5fe99e07fc757a58e2/measures/ResStockArguments/measure.rb#L838-L839)? Or maybe we should reconsider whether some of the BuildResHPXML arguments should be required.

For example, for options with `cooling_system_type=none`, we set `cooling_system_cooling_efficiency=auto`. Then at the place you linked to, we set `cooling_system_cooling_efficiency=0` (or perhaps to its default value). Something like that?

shorowit:
> For example, for options with `cooling_system_type=none`, we set `cooling_system_cooling_efficiency=auto`. Then at the place you linked to, we set `cooling_system_cooling_efficiency=0` (or perhaps to its default value). Something like that?

Yep, exactly. Maybe it would make sense to use 99999 just to ensure that the value isn't being accidentally used, as that would surely cause something to blow up. I don't know if that's a good idea or not.

> I suppose it's up to you whether we want to change, e.g., `cooling_system_cooling_efficiency` to be an optional argument.

I think I'm okay with doing this. Or at least trying. I'm not sure what the disadvantages are exactly. But it must be confusing to users to have to provide a `cooling_system_cooling_efficiency` value even when they have no cooling system, for example. And we could probably clean up some code in OS-HPXML (like [here](https://github.com/NREL/OpenStudio-HPXML/blob/9112f440b93ad3d3b6100d2d876164b617d1e2f8/workflow/hpxml_inputs.json#L97)) if we did this.

joseph-robertson:
Superseded by #1316.

PR #1323, opened 2024-11-18
Add misc appliances to panel load calc

lixiliu:
## Pull Request Description
Assign the following miscellaneous permanently connected appliances to homes for NEC panel load calculations:
- **Microwave**: 100% of homes, power level based on the number of occupants
    * Sizing (in cu ft) based on [Whirlpool's microwave buying guide](https://www.whirlpool.com/blog/kitchen/how-to-choose-a-microwave.html), then power is assigned to each cu ft class based on general search

      | Occupants | Size (cu ft) | Power (W) |
      | -- | -- | -- |
      | 0 - 2 | 0.5 - 0.9 | 900 |
      | 3 - 4 | 1.0 - 1.6 | 1100 |
      | 4+ | 1.7 - 2.2 | 1250 |

- **Garbage disposal**: 52% of homes (AHS 2013), power level based on the number of occupants
    * Power rating estimated based on average load amperage (x 120V), not horsepower rating as they appear to consume more power
    * Power rating based on InSinkErator product labels
    * Sizing based on Lowe's [garbage disposal buying guide](https://www.lowes.com/n/buying-guide/garbage-disposal-buying-guide)
 
      | Occupants | Size (HP) | Avg Load (A) | Power (W) |
      | -- | -- | -- | -- |
      | 0 - 1 | ⅓ | 5.6 | 672 |
      | 2 - 3 | ½ | 6.3 | 756 |
      | 4 - 5 | ¾ | 9.5 | 1140 |
      | 5+ | 1 | 10.2 | 1224 |

- **Garage door opener**: assign one ½ HP if home has a garage, regardless of the number of garage
    * Power rating converted from horsepower rating at 1 mech HP = 745.7 W

@afontani @JLReyna - should this be based on occupants or bedrooms, is there a conversion between the two?

### Related Pull Requests

### Related Issues

#### Required:
- [ ] [Technical reference documentation](https://github.com/NREL/resstock/tree/develop/docs/read_the_docs) has been updated


measures/ResStockArguments/measure.rb
```diff
@@ -844,14 +844,45 @@ def run(model, runner, user_arguments)
...
       args[:electric_panel_load_other_power] = 0
     end
-    microwave_power = [0, 100, 200, 300].sample(1, random: Random.new(args[:building_id]))
-    garbage_disposal_power = [0, 200, 400, 600].sample(1, random: Random.new(args[:building_id]))
-    args[:electric_panel_load_other_power] += microwave_power[0]
-    args[:electric_panel_load_other_power] += garbage_disposal_power[0]
+    # Assume all homes have a microwave
+    if args[:geometry_unit_num_occupants] <= 2
+      microwave_power = 900 # W, small, <= 0.9 cu ft
+    elsif args[:geometry_unit_num_occupants] <= 4
```
lixiliu, line 854: @joseph-robertson - is this okay or do I need to define the condition as `args[:geometry_unit_num_occupants] > 2 && args[:geometry_unit_num_occupants] <= 4`?

lixiliu: I think so. Microwave exists in unoccupied homes too.
I am also wondering if we should do this based on # of bedrooms instead. Is there a conversion we can use between occupants and bedrooms?

JLReyna:
@lixiliu  - in response to your question, the `Occupants.tsv` has a dependency on `Bedrooms.tsv`. I don't actually think we should have a dependency on either for garage doors. To have multiple openers (or differently sized openers), you'd be assuming some sort of factor driving the increase. I'd probably argue Geometry Floor Area if we were to choose something, but I don't think we know enough to develop that. Size of the garage (1 car, 2 car, 3 car) would be the clearest indicator, but even that doesn't mean increases in the opener. So long winded way to say I don't think this should have dependencies. @afontani do you agree?

lixiliu:
> @lixiliu - in response to your question, the `Occupants.tsv` has a dependency on `Bedrooms.tsv`. I don't actually think we should have a dependency on either for garage doors. To have multiple openers (or differently sized openers), you'd be assuming some sort of factor driving the increase. I'd probably argue Geometry Floor Area if we were to choose something, but I don't think we know enough to develop that. Size of the garage (1 car, 2 car, 3 car) would be the clearest indicator, but even that doesn't mean increases in the opener. So long winded way to say I don't think this should have dependencies. @afontani do you agree?

Thanks @JLReyna - and apologies for the confusion, the question is really for assigning microwave and garbage disposal, not garage door (only one rating is assigned if has garage). There are market guidance on their sizing based on household size, which I am using currently. But depending solely on occupants means we may have undersized equipment for vacant homes.

measures/ResStockArguments/measure.rb
```diff
@@ -860,14 +860,45 @@ def run(model, runner, user_arguments)
...
+    else
+      microwave_power = 1250 # W, large, 1.7-2.2 cu ft, 5+ ppl
+    end
+
+    garbage_disposal_ownership = 0.52 # AHS 2013
+    if Random.new(args[:building_id]).rand > garbage_disposal_ownership
+      garbage_disposal_power = 0
+    else
+      # Power estimated from avg load amp not HP rating, from InSinkErators
+      if args[:geometry_unit_num_bedrooms] <= 1
```
lixiliu, line 881: Note, this has been updated to vary based on bedrooms instead of occupants.

JLReyna (approved):

PR #1331, opened 2024-12-13
Load flexibility yingli

yingli-NREL:
## Pull Request Description
1. Revise the peak time input json files and related code in the measure, which get rid of the variable "on_peak_duration"
2. Add setpoint clip function to to prevent excessive heating or cooling
3. Revise  the default value of the load flexibility scenario design
4. Did a local test run, the run is successful
5. Add the test to test the run of the whole measure and setpoints are modified as designed


measures/ResStockArgumentsPostHPXML/resources/hvac_flexibility/setpoint_modifier.rb
```diff
@@ -47,27 +47,49 @@ def modify_setpoints(setpoints, flexibility_inputs)
...
+      return setpoint
+    end
+    if day_type == 'cooling'
+      if setpoint > 80
+        setpoint = 80
+      elsif setpoint < 60
+        setpoint = 60
+      end
+      return setpoint
+    end
```
rajeee: I think this can be simplified a bit to

```suggestion
return 80 if day_type == 'heating' && setpoint > 80
return 55 if day_type == 'heating' && setpoint < 55
return 80 if day_type == 'cooling' && setpoint > 80
return 60 if day_type == 'cooling' && setpoint < 60
setpoint
```

I think it's cleaner to have a flat else-if and early return when possible.

resources/options_lookup.tsv
rajeee, line 1: What changes are you making on this file? Is it required?

measures/ResStockArgumentsPostHPXML/resources/hvac_flexibility/setpoint_modifier.rb
yingli-NREL: sounds good!

resources/options_lookup.tsv
yingli-NREL: I deleted the loadflex_peak_duration_hours argument in the LoadFlexibility

yingli-NREL (commented):
addressed all comments

PR #1340, opened 2025-01-23
 break space update

yingli-NREL:
## Pull Request Description

1. update the breaker space distribution csv table
2. add building type and floor area as dependent variables


lixiliu (approved):
LGTM, thanks Yingli. Merge when all CI tests passed.

PR #1230, opened 2024-04-15
Run `run_analysis` in parallel

joseph-robertson:
## Pull Request Description

Up-to-date copy of https://github.com/NREL/resstock/pull/1131.
Closes #1130.

## Checklist

Not all may apply:

- [ ] ~Tests (and test files) have been updated~
- [x] Documentation has been updated
  - [ ] ~If related to resstock-estimation, checklist includes [data dictionary](https://github.com/NREL/resstock/tree/develop/resources/data/dictionary), [source report](https://github.com/NREL/resstock/tree/develop/project_national/resources/source_report.csv), [options saturation](https://github.com/NREL/resstock/tree/develop/project_national/resources/options_saturations.csv), [options_lookup](https://github.com/NREL/resstock/blob/develop/resources/options_lookup.tsv).~
- [x] Changelog has been updated
- [x] `openstudio tasks.rb update_measures` has been run
- [x] No unexpected regression test changes on CI (checked comparison artifacts)

PR #1342, opened 2025-01-28
Reference housing characteristics directly

rajeee:
## Pull Request Description
Remove a archaic workflow where we copy the resources and housing characteristics to the lib folder. Instead, pass the project_name and reference the housing characteristics directly.

### Related Pull Requests

### Related Issues

#### Required:


shorowit:
Do you plan to get rid of the lib folder creating/copying that run_analysis.rb does? 🙏

rajeee:
> Do you plan to get rid of the lib folder creating/copying that run_analysis.rb does? 🙏

Yes!

joseph-robertson:
You might take a look at https://github.com/NREL/resstock/pull/1230 -- I'm not sure how much overlap there is here.

rajeee:
> You might take a look at #1230 -- I'm not sure how much overlap there is here.

Thanks for linking it. I read through the issue and the implemented fix. This is what I gathered:

Before https://github.com/NREL/resstock/pull/1230 run_analyis would create lib folder and then copy housing_characterics and resources folder into it. It would also copy (if using precomputed) or sample the buildstock.csv file in the lib/housing_characterics directory. 
This would cause problem if someone wants to spin-off another ResStock simulation when one is already under way since the lib folder is wiped at the end (though can be avoided by using --debug) and the buildstock.csv will be overwritten by the newer run. The PR makes it so that the buildstock.csv is moved to the results_directory immediately from the  lib/housing_characterics if sampled and used from there, or used from the original location if using precomputed. It also makes smart decision about copying the lib folder and does so only if the files have changed. This allows running multiple ResStock simulations in multiple terminals, typically without problem, though, doing multiple simulations that require sampling has to be staggered enough so that to program don't attempt to write the buildstock.csv at once.

This PR takes it even further and gets rid of creating the lib folder altogether. The lib folder is an artifact of the way we used to run ResStock using PAT on cloud computing - but ResStock has since moved on to using BuildStockBatch long time ago. I think https://github.com/NREL/resstock/pull/1230 is a case of optimizing something that should have instead been removed!

measures/ApplyUpgrade/measure.rb
```diff
@@ -121,7 +127,7 @@ def run(model, runner, user_arguments)
     end
 
     upgrade_name = runner.getStringArgumentValue('upgrade_name', user_arguments)
-
+    args = runner.getArgumentValues(arguments(model), user_arguments)
```
joseph-robertson: If you're going to access new `:project_name` from the `args` hash, I'd suggest using `args[:upgrade_name]` -- therefore you can delete the `upgrade_name = ...` line.

resources/data/dictionary/inputs.csv
```diff
@@ -132,6 +132,7 @@ build_existing_model.orientation,:ref:`orientation`
 build_existing_model.overhangs,:ref:`overhangs`
 build_existing_model.plug_load_diversity,:ref:`plug_load_diversity`
 build_existing_model.plug_loads,:ref:`plug_loads`
+build_existing_model.project_name,The name of the housing characteristics project.
```
joseph-robertson: Put this a bit lower in the list so it's organized along with the other `build_existing_model.xxx` entries which are not actually tsv parameters?

joseph-robertson (commented):
Can you explain the new `project_name` arguments?

joseph-robertson:
> > You might take a look at #1230 -- I'm not sure how much overlap there is here.
> 
> Thanks for linking it. I read through the issue and the implemented fix. This is what I gathered:
> 
> Before #1230 run_analyis would create lib folder and then copy housing_characterics and resources folder into it. It would also copy (if using precomputed) or sample the buildstock.csv file in the lib/housing_characterics directory. This would cause problem if someone wants to spin-off another ResStock simulation when one is already under way since the lib folder is wiped at the end (though can be avoided by using --debug) and the buildstock.csv will be overwritten by the newer run. The PR makes it so that the buildstock.csv is moved to the results_directory immediately from the lib/housing_characterics if sampled and used from there, or used from the original location if using precomputed. It also makes smart decision about copying the lib folder and does so only if the files have changed. This allows running multiple ResStock simulations in multiple terminals, typically without problem, though, doing multiple simulations that require sampling has to be staggered enough so that to program don't attempt to write the buildstock.csv at once.
> 
> This PR takes it even further and gets rid of creating the lib folder altogether. The lib folder is an artifact of the way we used to run ResStock using PAT on cloud computing - but ResStock has since moved on to using BuildStockBatch long time ago. I think #1230 is a case of optimizing something that should have instead been removed!

So the feature that #1230 enabled is still available after the changes in this PR?

rajeee:
> Can you explain the new `project_name` arguments?

It's the project directory name - basically "project_testing" or "project_national".

rajeee:
> > > You might take a look at #1230 -- I'm not sure how much overlap there is here.
> > 
> > 
> > Thanks for linking it. I read through the issue and the implemented fix. This is what I gathered:
> > Before #1230 run_analyis would create lib folder and then copy housing_characterics and resources folder into it. It would also copy (if using precomputed) or sample the buildstock.csv file in the lib/housing_characterics directory. This would cause problem if someone wants to spin-off another ResStock simulation when one is already under way since the lib folder is wiped at the end (though can be avoided by using --debug) and the buildstock.csv will be overwritten by the newer run. The PR makes it so that the buildstock.csv is moved to the results_directory immediately from the lib/housing_characterics if sampled and used from there, or used from the original location if using precomputed. It also makes smart decision about copying the lib folder and does so only if the files have changed. This allows running multiple ResStock simulations in multiple terminals, typically without problem, though, doing multiple simulations that require sampling has to be staggered enough so that to program don't attempt to write the buildstock.csv at once.
> > This PR takes it even further and gets rid of creating the lib folder altogether. The lib folder is an artifact of the way we used to run ResStock using PAT on cloud computing - but ResStock has since moved on to using BuildStockBatch long time ago. I think #1230 is a case of optimizing something that should have instead been removed!
> 
> So the feature that #1230 enabled is still available after the changes in this PR?

Yes - it's even better now. In #1230,  buildstock.csv was written to lib/housing_characteristics first and then moved to the results_dir. In this PR, it's directly written to the results_dir.

joseph-robertson:
> > Can you explain the new `project_name` arguments?
> 
> It's the project directory name - basically "project_testing" or "project_national".

So it's, e.g., this? https://github.com/NREL/resstock/blob/no_lib/project_national/national_baseline.yml#L5

So why not call it `project_directory`?

rajeee:
> > > Can you explain the new `project_name` arguments?
> > 
> > 
> > It's the project directory name - basically "project_testing" or "project_national".
> 
> So it's, e.g., this? [`no_lib`/project_national/national_baseline.yml#L5](https://github.com/NREL/resstock/blob/no_lib/project_national/national_baseline.yml?rgh-link-date=2025-01-29T22%3A50%3A23Z#L5)
> 
> So why not call it `project_directory`?

Yes. It was an oversight. I will rename it to project_directory

measures/ApplyUpgrade/measure.rb
```diff
@@ -47,6 +47,11 @@ def arguments(model) # rubocop:disable Lint/UnusedMethodArgument
     upgrade_name.setDefaultValue('My Upgrade')
     args << upgrade_name
 
+    project_directory = OpenStudio::Measure::OSArgument::makeStringArgument('project_directory', true)
+    project_directory.setDisplayName('Project Directory')
+    project_directory.setDescription('The directory containing the housing characteristics folder (e.g., project_national).')
+    args << project_directory
```
joseph-robertson, line 53: Why is this new `project_directory` argument required, but the one in BuildExistingModel is optional?

workflow/run_analysis.rb
```diff
@@ -338,7 +324,9 @@ def run_workflow(yml, in_threads, measures_only, debug_arg, overwrite, building_
 
     if upgrade_name != 'Baseline'
       apply_upgrade_measure = { 'measure_dir_name' => 'ApplyUpgrade',
-                                'arguments' => { 'run_measure' => 1 } }
+                                'arguments' => { 'run_measure' => 1, 
+                                  'project_directory': project_directory
+                                  } }
```
joseph-robertson: Need to run `openstudio tasks.rb update_measures`.

workflow/run_analysis.rb
```diff
@@ -77,29 +77,14 @@ def run_workflow(yml, in_threads, measures_only, debug_arg, overwrite, building_
 
   Dir.mkdir(results_dir)
 
-  # Create lib folder
-  lib_dir = File.join(thisdir, '../lib')
-  resources_dir = File.join(thisdir, '../resources')
-  housing_characteristics_dir = File.join(buildstock_directory, project_directory, 'housing_characteristics')
```
joseph-robertson, line 83: Doesn't appear the `buildstock_directory` variable gets used anymore.

measures/ApplyUpgrade/measure.rb
rajeee: Both should be required. Let me change.

workflow/run_analysis.rb
rajeee: Yeah, it's not required anymore for run_analysis.

joseph-robertson (approved):
Assuming CI passes, I think this is ready to go.

PR #1369, opened 2025-03-27
Add Electric Vehicle Demand Flexibility Upgrades

rajeee:
## Pull Request Description
Add Electric Vehicle load flexibility upgrade measures to the SDR yaml.

### Related Pull Requests

### Related Issues

#### Required:
     - [x] [Technical development guide](https://github.com/NREL/resstock/tree/develop/docs/technical_development_guide)

  - [x] [options_lookup](https://github.com/NREL/resstock/blob/develop/resources/options_lookup.tsv)


measures/ResStockArgumentsPostHPXML/resources/ev_flexibility/ev_schedule_modifier.rb
```diff
@@ -0,0 +1,52 @@
...
+  require resource_file
+end
+
+
+class EVScheduleModifier < ScheduleModifier
+
+  # Modifies the EV schedule based on flexibility inputs
+  # @param ev_schedule [Array] Array of EV charging values (+ve is charging, -ve is discharging)
+  # @param flexibility_inputs [FlexibilityInputs] Inputs for schedule modification
+  # @return [Array] Modified EV schedule with charging (positive values) set to 0 during peak periods
```
aspeake: Returns a hash not array

measures/ResStockArgumentsPostHPXML/tests/test_resstock_arguments_post_hpxml.rb
```diff
@@ -17,25 +17,58 @@ def test_hvac_load_flexibility_measure
...
+    puts 'Testing EV Load Flexibility'
+    # Define test parameters
+    test_cases = [
+      { dst_enabled: false, name: 'without DST' },
+      { dst_enabled: true, name: 'with DST enabled' }
+    ]
+
+    curdir = File.dirname(__FILE__)
+    osw_hash_orgi = JSON.parse(File.read(File.join(curdir, 'test_template.osw')))
+    # find the index of the ResStockArgumentsPostHPXML measure in the steps and set ev_flex_enabled to false
```
aspeake: Incorrect comment? Not setting ev_flex to false

measures/ResStockArgumentsPostHPXML/tests/test_resstock_arguments_post_hpxml.rb
```diff
@@ -17,25 +17,58 @@ def test_hvac_load_flexibility_measure
...
+    measure_index = osw_hash_orgi['steps'].find_index { |step| step['measure_dir_name'] == 'ResStockArgumentsPostHPXML' }
+    osw_hash_orgi['steps'][measure_index]['arguments']['hvac_flex_peak_offset'] = 0
+    osw_hash_orgi['steps'][measure_index]['arguments']['hvac_flex_pre_peak_duration_hours'] = 0
```
aspeake, line 61: Would be good to have a test with both EV and HVAC

project_national/sdr_upgrades_tmy3.yml
```diff
@@ -1177,4 +1177,10 @@ upgrades:
...
     package_apply_logic: *has_cooling_or_heating
     options:
-      - option: HVACFlexibility|pre_only_1hr_2deg
\ No newline at end of file
+      - option: HVACFlexibility|pre_only_1hr_2deg
+
+  - upgrade_name: EV for all with load flexibility
+    options:
+      - option: Electric Vehicle Ownership|Yes
+      - option: Electric Vehicle Charger|Level 1 charger
```
aspeake, line 1198: I know this is not the current yml, but make sure we are setting L2 charger

resources/options_lookup.tsv
```diff
@@ -9245,6 +9245,7 @@ Duct Location	Heated Basement	ResStockArguments	ducts_supply_location=basement -
 Duct Location	Living Space	ResStockArguments	ducts_supply_location=conditioned space	ducts_supply_surface_area=auto	ducts_supply_surface_area_fraction=auto	ducts_return_location=conditioned space	ducts_return_surface_area=auto	ducts_return_surface_area_fraction=auto	ducts_number_of_return_registers=auto
 Duct Location	None	ResStockArguments	ducts_supply_location=conditioned space	ducts_supply_surface_area=auto	ducts_supply_surface_area_fraction=auto	ducts_return_location=conditioned space	ducts_return_surface_area=auto	ducts_return_surface_area_fraction=auto	ducts_number_of_return_registers=0
 Duct Location	Unheated Basement	ResStockArguments	ducts_supply_location=basement - unconditioned	ducts_supply_surface_area=auto	ducts_supply_surface_area_fraction=auto	ducts_return_location=basement - unconditioned	ducts_return_surface_area=auto	ducts_return_surface_area_fraction=auto	ducts_number_of_return_registers=auto
+EVFlexibility	enabled	ResStockArgumentsPostHPXML	ev_flex_enabled=true	flex_random_shift_minutes=30
```
aspeake: I think this could be risky to have two, non-mutually exclusive parameters (`EVFlexibility` and `HVACFlexibility`) change the same argument (`flex_random_shift_minutes`). It could be assigned twice if there is a measure combining DF. I'd recommend custom shift arguments for EVs and HVAC.

measures/ResStockArgumentsPostHPXML/measure.rb
```diff
@@ -98,13 +105,17 @@ def run(model, runner, user_arguments)
...
       hpxml_bldg = @hpxml.buildings[index]
-      if hpxml_bldg.hvac_controls.to_a.length == 0
-        runner.registerInfo("Skipping hvac flexibility for building #{index + 1} since it has no HVAC controls.")
-        next
+      if hpxml_bldg.hvac_controls.to_a.length != 0 && !skip_hvac_flexibility?(args)
+        hvac_schedule = create_hvac_schedule(index)
+        modified_schedule = modify_hvac_schedule(index, hvac_schedule)
+        write_schedule(modified_schedule, building, index, output_csv_path)
+      end
+      if args[:ev_flex_enabled] && !skip_ev_flexibility?(args)
```
aspeake: Does this not check the same thing twice?

resources/options_lookup.tsv
rajeee: I initially had two different flex_random_shift_minutes input for HVAC and EV flexibility but later converted to a single argument. This is because if a unit is responding to DF request, they will be shifting all of their load in the same way. In no scenario would the peak period for HVAC start at 4:30 but peak period for EV start at 5:30 - they would both either be 4:30, or 5:30. 

You raise a valid point that EVFlexibility and HVACFlexibility might assign this shared argument differently leading to a conflict - however I don't think the solution to this is using independent arguments for the reason I outlined above. I think we can just raise a warning about values being overridden as a safety check. But also, and @afontani can speak more to this, we are starting to think of ResStock as a single whole entity instead of pieces that individually has to be robust to all sorts of input. In that paradigm, as long as in the options_lookup.tsv all parameters pass same values to flex_random_shift_minutes, we are good. We don't need to consider all the what-ifs scenario of someone manually modifying the options-lookup.tsv.

measures/ResStockArgumentsPostHPXML/measure.rb
rajeee: Good catch!

resources/options_lookup.tsv
aspeake: I am trying to avoid a lot of changes that account for every possible future, for example, there could be "EV peaks" in the future, but since that requires new input files, we would make code changes at the same time. This one though could be easy to overlook with a future change, whether a warning or a different argument is the solution, @afontani can weigh in. Maybe there isn't an obvious scenario where we want different shifts, but seems bad practice to allow overwriting. Are we sure a warning/error doesn't already get thrown?

project_national/sdr_upgrades_tmy3.yml
```diff
@@ -1177,4 +1177,17 @@ upgrades:
...
+      - option: Electric Vehicle Ownership|Yes
+      - option: Electric Vehicle Charger|Level 2 charger
+      - option: EVFlexibility|enabled
+
+  - upgrade_name: Efficienct EV + L2 Charger + DF
+    options:
+      - option: Electric Vehicle Ownership|Yes
+      - option: Electric Vehicle Charger|Level 2 charger
+      - option: Electric Vehicle Efficiency Increase|15%
+      - option: EVFlexibility|enabled
```
aspeake: to-do: apply only to occupied units

afontani: [b03146c](https://github.com/NREL/resstock/pull/1369/commits/b03146c9c6ea26f85d1ec26c3fc80abf8f6ced38)

resources/options_lookup.tsv
afontani: I am moving all the demand flexibility arguments assigned in `ResStockArgumentsPostHPXML` to the same area of `options_lookup.tsv` and breaking out the random shift into its own parameter and option sets.

project_national/sdr_upgrades_tmy3.yml
afontani: Oops. I missed this too in the other PR.

afontani: EVs are only being given to units without an EV. If there is no EV, the Charger is "Option=None". Therefore, no L2 chargers in the baseline are being downgraded.

measures/ResStockArgumentsPostHPXML/tests/test_resstock_arguments_post_hpxml.rb
afontani: @rajeee: I tried a bit, but I think you would be faster at this than I.

afontani: addressed: [b96bb7d](https://github.com/NREL/resstock/pull/1369/commits/b96bb7d61c35cafbd2602d0f8b7f839831f10b09) and [f5038a5](https://github.com/NREL/resstock/pull/1369/commits/f5038a55c75971ec5c58acf86b933255b59baf7c)

PR #1384, opened 2025-04-17
Add SDR options analysis and minimal buildstock.csv through CI

rajeee:
## Pull Request Description
Run the upgrades analyzer to generate the options apply logic report for the sdr yaml. These report helps catch and correct apply_logic errors, and also can be used in documentation to explain exactly how many building models the various pieces of the logic applies to.

In addition, it also generates a minimal version of buildstock.csv file. This file contains the minimal number of buildings but with enough diversity such that all of the options in the sdr yaml will apply to at least one building (when possible*).

All these files are commited to project_national/resources/ folder. 

*If the logic applies to at least one building in 550K sample. 

### Related Pull Requests

### Related Issues

#### Required:
- [x] Add to the [changelog_dev.rst file](https://github.com/NREL/resstock/blob/develop/docs/technical_development_guide/source/changelog/changelog_dev.rst)
- [x] No unexpected regression test changes on CI (comparison artifacts have been checked)


afontani:
The new CI workflow. We probably need to make update-results dependent on `sdr_options_analysis`. I would maybe just call the job `sdr_upgrade_analysis`.
Screenshot 2025-04-17 at 6 36 04 PM

afontani:
@rajeee: I guess we could pull out the sdr tests and put them in the new job. I like this new job split, because the integration and analysis tests are more about the baseline and code changes, where the sdr job is about upgrades. These jobs are also parallel to each other which is also good.

.github/workflows/config.yml
```diff
@@ -88,6 +92,58 @@ jobs:
...
+          sudo -E apt-get install -y python3-pip
+
+      - name: Install python dependencies
+        run: |
+          python3 -m pip install --progress-bar off --upgrade pip
+          pip3 install --progress-bar off pandas pyyaml
+
+      - name: Install buildstock_query
+        run: |
+          python3 -m pip install git+https://github.com/NREL/buildstock-query.git@minimal_buildings
```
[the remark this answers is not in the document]
afontani: We just need to make this change when the minimal_buildings feature is merged in buildstock-query.

project_national/resources/sdr_option_application_detailed_report.txt
afontani, line 1: Can you remind me about the usefulness of this file? The applicability is in `sdr_option_application_report.csv` file. What additional information is in this file? It is long, and I would like it to be useful if we keep it. Is there another form that this could take that is more readable?

project_national/resources/sdr_option_application_report.csv
afontani, line 1: I would like to have a test that errors out if there is 0 applicability. Example: https://github.com/NREL/resstock/blob/278e8eeb161cfed69e2d6bd6e91fcb6aded18004/project_national/resources/sdr_option_application_report.csv#L48

afontani (changes requested):
Overall, I think this will work. Since you are pip installing buildstock-query into the CI environment, this is cleaner than I expected. I added some comments, but we can move forward with the development.

project_national/resources/sdr_option_application_detailed_report.txt
rajeee: This provides a more detailed x-ray into the apply logic. For example, the .csv file[ in this line ](https://github.com/NREL/resstock/blob/278e8eeb161cfed69e2d6bd6e91fcb6aded18004/project_national/resources/sdr_option_application_report.csv#L189) tells you that the option applied to 16.7% of the buildings, which can seem correct. But if you look at the [detailed report txt file for this option](https://github.com/NREL/resstock/blob/278e8eeb161cfed69e2d6bd6e91fcb6aded18004/project_national/resources/sdr_option_application_detailed_report.txt#L6324), you will find that that the "Water Heater Efficiency|Electric Heat Pump, 66 gal, 3.35 UEF" logic had no bearing on this logic since there is not water heater with that specification in the 550K sample. This may or may not be correct/expected, but the report file provides you the information to dig further.

Regarding alternative forms, few ideas come to mind.
1. Split them into one file per upgrade. 
   Although the file is large, it is purely text so github seems to be able to handle it okay. When a PR is made with changes/addition to just 1 upgrade, the diff view will only show section relevant to that upgrade (for example [the diff](https://github.com/NREL/resstock/pull/1384/commits/2d2198153039f35c1179b498a3f8e1af0cfb1c05#diff-46f935f57520951706f53e465cc8473bb2105987aa956328ded5203eadd6379f) after you pulled in the EV DF) so it shouldn't be that verbose in such cases. So, I am not too sure if splitting is worthwhile.
   
  2. This file would actually be of about the same length as the sdr yaml (minus the comments in the yaml, plus extra filler description in the report txt) if not for a detailed [options-interplay-report](https://github.com/NREL/resstock/blob/278e8eeb161cfed69e2d6bd6e91fcb6aded18004/project_national/resources/sdr_option_application_detailed_report.txt#L384-L405). This is a more confusing part, and we could maybe remove this, but i think it is useful. For example, from this report, we can tell that 74% of buildings got exactly 3 options applied and 22% got exactly 9 options applied, which serves as a nice verification that options are being applied as intended (like no buildings that got 6 options applied etc).
  
image

I understand that this report file feels overwhelming, but once we start getting PR with one upgrade at a time, it should be manageable.

project_national/resources/sdr_option_application_report.csv
rajeee: How about erroring out if any apply_logic block (or sub-block) applies to 100% of the buildings or 0% of the buildings? 
I am a bit hesitant to portray that as a logic error though since the 0%/100% might be a sample size issue and not necessarily a logic issue. Nevertheless, if we are going to do the 550K sample, those blocks are useless and should probably removed. Thoughts?

.github/workflows/config.yml
```diff
@@ -88,6 +92,94 @@ jobs:
...
+          rm national_550ksamples.csv
+
+      - name: Upload SDR options analysis artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: sdr_options_analysis
+          path: |
+            project_national/resources/sdr_minimal_buildstock.csv
+            project_national/resources/sdr_option_application_detailed_report.txt
+            project_national/resources/sdr_option_application_report.csv
```
afontani, line 142: Do we need to upload the report files as artifacts? You currently have them committed. We should upload the minimal buildstock.csv in L140.

.github/workflows/config.yml
```diff
@@ -88,6 +92,94 @@ jobs:
...
+          retention-days: 90
```
afontani, line 143:
```suggestion
          retention-days: 90
          if-no-files-found: error          
```

afontani (changes requested):
Add a changelog entry.

.github/workflows/config.yml
rajeee: Yes - because if the CI fails because some apply logic applies to zero buildings, we won't commit the files. In such cases, it will be useful to be able to download the artifact and debug.

afontani: Oh! Good point.

.gitattributes
```diff
@@ -1,6 +1,7 @@
 # Declare files that will always have CRLF line endings on checkout.
 
 project_*/housing_characteristics/*.tsv text eol=crlf
+project_*/resources/*.csv text eol=crlf
```
afontani, line 4: This is a good addition

afontani (approved):

PR #1197, opened 2024-01-24
Heat pump controls, try 2

joseph-robertson:
## Pull Request Description

Addresses #1154.
Tests https://github.com/NREL/OpenStudio-HPXML/pull/1584.
Supersedes #1179.

## Checklist

Not all may apply:

- [ ] Tests (and test files) have been updated
- [ ] Documentation has been updated
  - [ ] If related to resstock-estimation, checklist includes [data dictionary](https://github.com/NREL/resstock/tree/develop/resources/data/dictionary), [source report](https://github.com/NREL/resstock/tree/develop/project_national/resources/source_report.csv), [options saturation](https://github.com/NREL/resstock/tree/develop/project_national/resources/options_saturations.csv), [options_lookup](https://github.com/NREL/resstock/blob/develop/resources/options_lookup.tsv).
- [ ] Changelog has been updated
- [ ] `openstudio tasks.rb update_measures` has been run
- [ ] No unexpected regression test changes on CI (checked comparison artifacts)

PR #1408, opened 2025-05-13
Use Autosizing Limits and Maintain Duct System Curve

### LINKED ISSUE #1154: Account for baseline house duct size in the energy model
The current approach for modeling fan power effectively results in upgrades having a higher power use for a given flow despite a targeted fan efficiency of 0.375 W/cfm in the upgraded home compared to 0.5 W/cfm in the baseline home. Comparing pressure rise vs flow curves shows that the fan pressure rise has increased for a given flow for the upgrade.  Essentially the duct system has become more restrictive. 

The first step is to expose [fan W/cfm in OS-HPXML](https://github.com/NREL/OpenStudio-HPXML/issues/1530).

joseph-robertson:
## Pull Request Description

Closes #1154. Implements pieces of https://github.com/NREL/resstock/pull/1197. Uses https://github.com/NREL/OpenStudio-HPXML/pull/1584.

ResStock currently assumes the existing home's duct system for HVAC equipment upgrades. Without restricting the upgraded equipment's capacity to the maximum of the existing duct system's heating/cooling airflow rate divided by **400 cfm/ton**, fan pressure rise increases and results in higher power use. This is avoided by limiting the upgraded equipment's capacity based on the existing duct system.

Additionally, the blower fan efficiency (W/cfm) is adjusted based on baseline model inputs to assume the duct system curve does not change pre- and post- heat pump installation.

The "duct restriction" feature is probably more appropiate for Furnace -> HP upgrades, and less so for other types of upgrades (e.g., Furnace -> Furnace, HP -> Furnace, or even HP -> HP). Although this PR does not strictly prevent applying the duct restriction to these types of upgrades, (a) informative statements are registered containing autosizing limit, airflow cfm, and fan efficiency adjustment values, and (b) the new argument name suggests its applicability for heat pumps.

Specifically, a new boolean argument `heat_pump_sizing_is_duct_limited` is introduced in the ResStockArguments measure. When set to `true`, the "duct restriction" feature is enabled. This PR also introduces new ASHP options "ASHP, SEER 16, 9.2 HSPF, Duct Limited" and "ASHP, SEER 24, 13 HSPF, Max Load, Duct Limited" for demonstrating use of the new argument. These new options are _not_ added to the SDR upgrades yml file.

## Checklist
- [x] Use of 400 cfm/ton; for HP to HP upgrades, a better assumption may be to use the same cfm/ton as the baseline so the capacities are the same. I think this means that we shouldn't limit the upgraded HP's capacity when the baseline system is a HP.
- [x] Jon doesn't think this duct restriction approach makes sense for upgrading HVAC equipment types other than HPs. "When upgrading to HP, especially with Max Load sizing, the upgraded equipment is likely to have a higher CFM than the baseline equipment. I don't think this is an issue for a furnace upgrade."
- [x] If we cannot come up with a more suitable cfm/ton value to divide by than 400 for, e.g., furnace upgrades, then replace the new `hvac_distribution_use_duct_restriction` argument name with something more specific to heat pumps . This argument would then be assigned using ASHP options for `HVAC Heating Efficiency`.
- [x] Update documentation
  - TDG. Heat Pump Upgrades rst docs page. "Duct Restriction" is not technically a 4th sizing methodology option; it can be optionally applied when using either ACCA or MaxLoad sizing methodology options.
  - TRG. 3b_ResStockInputs_HVAC.tex.
- [x] Leave the new duct limited options out of the SDR yml completely for now.
- [x] Phil to make a resstock issue about furnace oversizing. https://github.com/NREL/resstock/issues/1426

## Comparison

**ENERGY STAR heat pump with elec backup**
n_datapoints=100

image
image
image
image
image
image
image
image
image
image
image
image
image

### Related Pull Requests

### Related Issues

#### Required:
- [x] Add to the [changelog_dev.rst file](https://github.com/NREL/resstock/blob/develop/docs/technical_development_guide/source/changelog/changelog_dev.rst)
- [x] No unexpected regression test changes on CI (comparison artifacts have been checked)
- [x] Update documentation (one is required)
     - [x] [Technical reference guide](https://github.com/NREL/resstock/tree/develop/docs/technical_reference_guide)
     - [x] [Technical development guide](https://github.com/NREL/resstock/tree/develop/docs/technical_development_guide)


afontani:
For future consideration: Some new changes in OS-HPXML are coming with the new version of OpenStudio 3.10. Example 400 cfm/ton is changing and could be a more dynamic variable and not constant. Probably need some resolution when this goes into "develop" and the "latest-os-hpxml" branch with the new features.

afontani (approved):

PR #1458, opened 2025-07-29
Fix possible non deterministic surface area in InternalMass object

joseph-robertson:
## Pull Request Description

Ports over changes in https://github.com/NREL/OpenStudio-HPXML/pull/2048 ahead of merging `latest-os-hpxml`. This includes direct manual edits to `resources/hpxml-measures/HPXMLtoOpenStudio/resources/constructions.rb` which will be overwritten in a subsequent update (i.e., git subtree) to this file.

### Related Pull Requests

### Related Issues

#### Required:
- [ ] ~Add to the [changelog_dev.rst file](https://github.com/NREL/resstock/blob/develop/docs/technical_development_guide/source/changelog/changelog_dev.rst)~
- [ ] ~Update documentation (one is required)~
     - [ ] ~[Technical reference guide](https://github.com/NREL/resstock/tree/develop/docs/technical_reference_guide)~
     - [ ] ~[Technical development guide](https://github.com/NREL/resstock/tree/develop/docs/technical_development_guide)~

- [ ] ~Supporting resource files have been updated (related to resstock-estimation)~
  - [ ] ~[data dictionary](https://github.com/NREL/resstock/tree/develop/resources/data/dictionary)~
  - [ ] ~[source report](https://github.com/NREL/resstock/tree/develop/project_national/resources/source_report.csv)~
  - [ ] ~[options saturation](https://github.com/NREL/resstock/tree/develop/project_national/resources/options_saturations.csv)~
  - [ ] ~[options_lookup](https://github.com/NREL/resstock/blob/develop/resources/options_lookup.tsv)~


rajeee (approved):
image

I think this will work for now. I printed the area (before rounding) a couple of times and it changes between 63.09974476799996 and 63.099744767999944. So, round(5) will safely produce stable result in this case. If somehow the flip was something between 679.19999499999999 and  679.199995000001 then round(5) would still result in two flipping numbers - 57.19999 and 679.2. We are just trusting that the instability is not at the 5th significant digit that we are rounding to. Which is the case for now - for this particular building - but isn't a fool proof solution. Hopefully, the openstudio bug gets fixed by the time we hit another lottery of the number sitting at the boundary of 5th significant digit!

PR #1467, opened 2025-08-12
Gracefully handle failures in upgrades

rajeee:
## Pull Request Description
Currently, if there are failures in any building in any upgrades, we remove those buildings from all upgrades. This was fine when the failure rate was extremely small, but when failure is substantial, this unnecessarily removes perfectly good buildings from other upgrades just because of one "bad" upgrade. This PR changes the logic so that only the failures in baseline is removed from all upgrades. Failures in upgrades are treated as if the upgrade didn't apply, and we copy baseline results for that building. This approach also aligns with what is done in ComStock.

### Related Pull Requests
https://github.com/NREL/buildstockbatch/pull/513

### Related Issues

#### Required:
- [ ] Add to the [changelog_dev.rst file](https://github.com/NREL/resstock/blob/develop/docs/technical_development_guide/source/changelog/changelog_dev.rst)


.github/workflows/config.yml
```diff
@@ -20,7 +20,7 @@ env:
   OPENSTUDIO_PLATFORM: Ubuntu-22.04-x86_64
   OPENSTUDIO_EXT: deb
   OPENSTUDIO_URL: https://github.com/NREL/OpenStudio/releases/download/v3.9.0
-  BUILDSTOCK_BRANCH: develop
+  BUILDSTOCK_BRANCH: pp_fix
```
asparke2, line 23: Does this need to get changed back before merging?

postprocessing/resstockpostproc/process_bsb_results.py
```diff
@@ -0,0 +1,82 @@
...
+        sys.exit(1)
+    if len(baseline_files) > 1:
+        print("Error: More than one baseline file found")
+        sys.exit(1)
+
+    baseline_file = baseline_files[0]
+    print(f"Processing baseline file: {baseline_file}")
+    baseline_df = pl.scan_parquet(baseline_file)
+    failed_bldgs = (
+        baseline_df.filter(pl.col("completed_status") == "Fail")
```
asparke2, line 37: Does ResStock ever have situations where the status is blank or "Success" but there are no numbers in the results data? ComStock has weird failure modes sometimes that we have to catch. If you haven't seen that probably don't need to worry about it.

postprocessing/resstockpostproc/process_metadata.py
```diff
@@ -9,62 +9,96 @@ def get_failed_bldgs(metadata_df: pl.LazyFrame) -> set[int]:
...
     missing_bldgs_df = missing_bldgs_df.with_columns([
         pl.lit(False).alias("applicability"),
         pl.lit(upgrade_num).alias("upgrade"),
     ]).drop("upgrade_name")
-    upgrade_cols = upgrade.collect_schema().names()
+    upgrade_cols = upgrade_df.collect_schema().names()
     missing_bldgs_df = missing_bldgs_df.join(upgrade_name_df,  how="cross")
-    upgrade = pl.concat([upgrade, missing_bldgs_df], how="diagonal_relaxed")
-    upgrade = upgrade.sort("bldg_id")
-    upgrade = add_saving_cols(upgrade, base)
```
asparke2, line 64: I think this needs to be called on the baseline DF as well so that that schemas for baseline and upgrade files are the same. Even though the savings will be zero, this is easier for users.

asparke2 (approved):

.github/workflows/config.yml
rajeee: Yes, will switch these back before merging.

postprocessing/resstockpostproc/process_bsb_results.py
rajeee: Yeah, those three are the only values I have seen. For the GHP upgrade with lots of failures (and possible weird corner cases), that's all we have.

image

